PR #6, opened 2019-12-17
feat: Adds an exponential backoff when failing

edas:
Before this change, if the server does not accept our modifications, we try to send them in a loop, until success. We noticed that if the server had a bug, this would flood the stack with unstoppable requests.

With this change, we implement an exponential backoff. We wait 128ms after the first failure, then double this amount each time until we succeed (and go back to 0ms). There is a maximum waiting time of 5 minutes (whatever the failure is, we will retry every 5 minutes)

This change also implement a catchup. After 4 failures in a row in sending new data to the server, we try to manually fetch changes from the server. This may resolve some out of sync situations.

### DIFF
--- a/src/lib/collab/channel.js
+++ b/src/lib/collab/channel.js
@@ -1,11 +1,162 @@
 import { EventEmitter2 } from 'eventemitter2'
 import { getVersion, sendableSteps } from 'prosemirror-collab'
 
+const minimumBackoff = 128 // 128ms
+const maximumBackoff = 1000 * 60 * 5 // Max 5 minutes
+const failuresBeforeCatchup = 4
+
 export class Channel {
   constructor(config, serviceClient) {
     this.config = config
     this.service = serviceClient
     this.eventEmitter = new EventEmitter2()
+    this.resetBackoff()
+    this.initializeStepsQueue()
+    this.isSending = false
+  }
+
+  /**
+   * Reset the backoff
+   */
+  resetBackoff() {
+    this.backoff = 0
+    this.failures = 0
+  }
+
+  /**
+   * Increase the backoff
+   */
+  increaseBackoff() {
+    this.failures = this.failures + 1
+    if (this.failures % failuresBeforeCatchup == 0) {
+      this.emit('needcatchup', { failures: this.failures })
+    }
+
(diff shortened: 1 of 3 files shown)


src/lib/collab/channel.js
```diff
@@ -1,76 +1,180 @@
...
     }
+  }
+
+  /**
+   * Program a push for new steps to the service
+   * @param {Function} getState - function to get a current proseMirror state
+   * @param {Object} state - proseMirror state
+   * @param {Object[]} localSteps - local steps to send
+   */
+  async sendSteps(getState, state, localSteps) {
```
Crash--: Are we using this `localSteps` attribute somewhere?

src/lib/collab/channel.js
```diff
@@ -1,76 +1,180 @@
...
+const failuresBeforeCatchup = 4
+
 export class Channel {
   constructor(config, serviceClient) {
     this.config = config
     this.service = serviceClient
     this.eventEmitter = new EventEmitter2()
+    this.resetBackoff()
+    this.initializeStepsQueue()
+    this.isSending = undefined
```
Crash--: Since this var starts with an `is` I expect that the value is `true` or `false` but not undefined

src/lib/collab/channel.js
edas: In `dequeueSteps`, requested by `processQueue`

src/lib/collab/channel.js
edas: I was playing with the face that undefined is indeed a false value but still not a value defined by the code below.

Agree that this difference is useless and may introduce questions. Will fix.

ptbrowne:
After the rush to launch cozy-notes, it would be good to have tests for this :)

Crash--:
>  it would be good to have tests for this :)

 it would be good to have tests :) 

:D

src/lib/collab/channel.js
edas: Documentation added

edas:
> After the rush to launch cozy-notes, it would be good to have tests for this :)

Oh yes, even if most is boiler plate or simple transform, and the rest is concurrent access that is hard/long to test.

PR #7, opened 2019-12-17
Add I18N

edas:
* Adds I18N for the Atlaskit Editor.

I added only FR and EN. If we want to benefit for all their locales, we should implement some kind of dynamic import as embedding everything would be too heavy

* Adds translations for our own components

Still, only FR and EN, no Transifex

* Cleanup for linter

### DIFF
--- a/src/locales/fr.json
+++ b/src/locales/fr.json
@@ -1,0 +1,20 @@
+{
+  "Notes": {
+    "Add": {
+      "add_note": "Ajouter une note"
+    },
+    "EditorLoading": {
+      "back_to_list": "Retour à la liste"
+    },
+    "EditorView": {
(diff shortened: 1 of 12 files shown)


src/locales/fr.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "EditorView": {
+      "main_placeholder": "Écrivez ce que vous souhaitez…"
+    },
+    "Editor": {
+      "title_placeholder": "Ici votre titre"
+    }
+  },
+  "Error": {
+    "unshared_title": "The partage n'est plus actif",
```
nono: s/The/Le/

edas: https://github.com/cozy/cozy-notes/pull/10

PR #8, opened 2019-12-17
CI setup

y-lohse:
Build, deploy and publish the app through travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,7 @@ env:
 cache:
   yarn: true
   directories:
-  - node_modules
-before_install:
-- curl -fsSL https://bootstrap.pypa.io/get-pip.py | python - --user
-- travis_retry pip install --user transifex-client==0.12.5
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,57 +2,36 @@ language: node_js
...
+script:
+  - yarn build
+before_deploy:
+  - yarn add cozy-app-publish
+deploy:
+  - provider: script
+    repo: cozy/cozy-notes
+    skip-cleanup: true
+    # deploy the build on a build branch and publish to the Cozy registry
+    script: DEPLOY_BRANCH=build && yarn deploy && yarn cozyPublish
```
Crash--: it means that if we already have a build in the build branch, this build will be erase.

Should we use something else? Like DEPLOY_BRANCH='auto-build' or `release`?

y-lohse: Yep, no problem to change it 👍

PR #9, opened 2019-12-17
fix: Correctly load cozy-ui

edas:
The app name in the Cozy-bar was at 11px instead of 1rem. This is the default for a browser and should have been overrided by cozy-ui.

Cozy-ui main stylesheet was missing. Fixed

### DIFF
--- a/src/styles/index.styl
+++ b/src/styles/index.styl
@@ -1,6 +1,12 @@
-@import '../../node_modules/cozy-ui/transpiled/react/stylesheet.css'
+@import '~cozy-ui/transpiled/react/stylesheet.css'
 
 // @stylint off
+
+/* Temporary fix waiting for https://github.com/cozy/cozy-bar/pull/675 */
+.coz-nav-apps-btns-main {
+    font-size: 1rem;
+}
+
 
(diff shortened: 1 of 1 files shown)


src/styles/index.styl
```diff
@@ -1,4 +1,5 @@
-@import '../../node_modules/cozy-ui/transpiled/react/stylesheet.css'
+@import '~cozy-ui/dist/cozy-ui.min.css'
+@import '~cozy-ui/transpiled/react/stylesheet.css'
```
y-lohse: Nit: [The official docs](https://github.com/cozy/cozy-ui#as-a-components-library) advocates to import this file in a jsx file (presumably the bootstraping file), since it contains global variables and nothing else that should be used in the app's css.

edas:
As you told me that cozy-ui.min.css was not needed, I changed and released a quick fix only for this instead of loading the whole cozy-ui css.

src/styles/index.styl
edas: What is the difference ? 
It seems better to me to have all CSS in one place

PR #11, opened 2019-12-18
Feat/home page

Crash--:
This PR changes the style of the Home Page of the app 

Since screenshots, I added a `u-mt-1` to the Add button 

Desktop
Capture d’écran 2019-12-18 à 11 18 50
Tablette 
Capture d’écran 2019-12-18 à 11 18 56
Mobile
Capture d’écran 2019-12-18 à 11 19 00


src/components/app.jsx
```diff
@@ -56,23 +62,37 @@ const PublicContext = withClient(({ client }) => {
...
-      </Main>
-      <IconSprite />
-    </Layout>
-  </HashRouter>
-)
+const App = ({ isPublic, breakpoints: { isMobile }, client }) => {
+  let appName = ''
+  if (isMobile) {
+    const data = client.getInstanceOptions()
+    appName = getDataOrDefault(data.cozyAppName, manifest.name)
```
y-lohse: I think we can hardcode the fallback name, no need to load the manifest.

src/components/notes/list.jsx
```diff
@@ -1,116 +1,75 @@
...
-const Item = props => (
-  <div className="note-item">
-    <Icon icon={icon} width={32} height={32} className="note-icon" />
-    <Link to={`/n/${props.note.id}`} className="note-link">
-      <ListItemText
-        primaryText={titleWithDefault(props.note)}
-        secondaryText="/Notes/2019/demo"
-      />
-    </Link>
+import icon from '../../assets/icons/icon_note_empty.svg'
```
y-lohse: Out of scope for this PR I guess, but shouldn't we use absolute pathes that can be overriden with webpack aliases?

src/components/notes/list.jsx
```diff
@@ -1,116 +1,75 @@
...
-        right={<Add />}
-      />
-    </>
-  )
-}
-
-export default ListHeader
+))
+/* const cellStyles = { flexGrow: 1 }
+const nameCellStyles = { flexGrow: 10 } */
```
y-lohse: Is this needed?

src/components/notes/list.jsx
```diff
@@ -1,116 +1,75 @@
...
+  </Table>
+))
+
+const TitleApp = translate()(({ t }) => (
+  <MainTitle className="u-pl-1">{t('Notes.List.my_notes')}</MainTitle>
+))
+
+const EmptyPage = withBreakpoints()(({ breakpoints: { isMobile } }) => (
+  //On Mobile, we don't want a margin top since we don't have title and
+  //we want the component to start at the end of the bar
```
y-lohse: The comment is confusing me, why do we not have a title on mobile? `<TitleApp />` is right there and I don't see a condition hiding it on mobile?

src/components/notes/list.jsx
```diff
@@ -1,116 +1,75 @@
...
+        <TableCell className="tableCellEmpty">
+          <EmptyComponent />
+        </TableCell>
+      </TableRow>
+    </TableBody>
+  </Table>
+))
+
+const TitleApp = translate()(({ t }) => (
+  <MainTitle className="u-pl-1">{t('Notes.List.my_notes')}</MainTitle>
```
y-lohse: Why do we need the left padding here?

src/components/notes/list.styl
```diff
@@ -0,0 +1,27 @@
+.empty {
```
y-lohse: [Our guidelines](https://docs.cozy.io/cozy-ui/styleguide/) say we don't use `{}`

src/components/notes/list.styl
```diff
@@ -0,0 +1,27 @@
...
+    height: 500px
+    display: flex
+}
+
+.tableCell {
+    flex-grow 1
+}
+
+.tableCellName {
+    flex-grow 10
```
y-lohse: Why 10..?

src/components/notes/list.jsx
Crash--: Because we want to have the title aligned with column's name and I don't know how to do other way? 
Capture d’écran 2019-12-18 à 11 58 33

src/components/notes/list.jsx
Crash--: TitleApp is not displayed on Mobile. You right, this `mt` should be on TitleApp and not on Stack

src/components/notes/list.styl
Crash--: because I had 14 columns in my head 🤔

src/components/notes/list.jsx
y-lohse: That's a correct way to do it imo. It's just that the relationship between the 2 is not explicit, maybe a comment would be good, but it's a nit.

Crash--: Let's see: 

I'll create: 
var PADDING_TABLE_FIRST_COLUMN = 1rem

and use it in a class. It'll be clearer :)

src/components/notes/list.jsx
Crash--: Done thanks

src/components/notes/list.jsx
Crash--: fixed

src/components/app.jsx
Crash--: This is already done like that in the browser entry point. What is the problem of importing the manifest since it is already imported and parsed

y-lohse: More code! But yeah I said this before seeing that the manifest was already imported elsewhere, so not really a problem.

PR #15, opened 2019-12-18
chore: Build without "v" prefix on tags

y-lohse:
We want to build tags even if they don't start with a `v`.


Crash--:
Are we sure about that? On drive we use vXXX, on Banks also

y-lohse:
On home and contacts we don't, without the `v` is becoming the new standard. It's more drive and banks that would need an update.

PR #17, opened 2019-12-18
feat: Add link to the Notes folder on Cozy Drive

Crash--:
This PR adds a link to the `Notes` folder created by the stack. We use the `referenced_folder` to fetch the folder. 

If the folder is not found (ATM we have a bug in the stack https://github.com/cozy/cozy-stack/pull/2287) we fallback to the Drive URL. 

I used markdown to put the URL in the message. But it comes with a drawback: we can't style the link... WDYT? 

Capture d’écran 2019-12-18 à 17 54 48

### DIFF
--- a/src/components/notes/hooks/useReferencedFolderForNote.jsx
+++ b/src/components/notes/hooks/useReferencedFolderForNote.jsx
@@ -1,0 +1,35 @@
+import { useState, useEffect } from 'react'
+import { CozyFolder } from 'cozy-doctypes'
+import { getFullLink } from 'lib/utils'
+
+const notesFolderDocument = {
+  _type: 'io.cozy.apps',
+  _id: 'io.cozy.apps/notes'
+}
+
+const useReferencedFolderForNote = client => {
+  const [notesFolder, setNotesFolder] = useState(null)
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const driveUrl = getFullLink(client)
+      try {
+        setNotesFolder(driveUrl)
+        const referencedFolder = await CozyFolder.getReferencedFolders(
+          notesFolderDocument
+        )
+        const folderUrl = getFullLink(client, referencedFolder[0]._id)
+        setNotesFolder(folderUrl)
+      } catch (error) {
+        setNotesFolder(driveUrl)
+      }
+    }
+    fetchData()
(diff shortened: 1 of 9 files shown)


src/components/notes/list.jsx
```diff
@@ -11,25 +12,43 @@ import {
...
-      id="empty"
-      icon={icon}
-      title={t('Notes.Empty.welcome')}
-      text={t('Notes.Empty.after_created')}
-    >
-      <Add className="u-mt-1" />
-    </Empty>
-  </div>
-))
+const EmptyComponent = translate()(
```
y-lohse: We could use this PR to extract this component into a separate file :)

src/components/notes/hooks/useReferencedFolderForNote.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const driveUrl = getFullLink(client)
+      try {
+        setNotesFolder(driveUrl)
+        const referencedFolder = await CozyFolder.getReferencedFolders(
+          notesFolderDocument
+        )
+        const folderUrl = getFullLink(client, referencedFolder[0]._id)
```
y-lohse: I know `getFullLink ` was here before, but seeing it in this context I realize that the name is not very clear. It's really not obvious that it generates a link to the drive app :/. I think renaming it would be a good idea.

src/components/notes/list.jsx
```diff
@@ -11,25 +12,43 @@ import {
...
+  withClient(({ t, client }) => {
+    const { notesFolder } = useReferencedFolderForNote(client)
+    return (
+      <div className="empty">
```
edas: nit: Not a priority today, but I see this component (large image then button) in multiple apps, and it may be a good idea to have it in cozy-ui in the future

y-lohse: [It exists](https://docs.cozy.io/cozy-ui/react/#!/Empty), it's what we're using here. But because of the markdown here there's some extra complexity.

Crash--: I think @edas talks about the need of the 500px height and translateZ etc. It should be part of the default component since we use that every time

PR #27, opened 2019-12-19
feat: Have the title on the tab bar of the browser

edas:
The browser tab is now named with the title of the document. This will help the awesome bar of Firefox and Chrome, and will allow user to put the page in their favorites

### DIFF
--- a/src/components/notes/editor.jsx
+++ b/src/components/notes/editor.jsx
@@ -8,9 +8,18 @@ import EditorLoading from './editor-loading'
 import CollabProvider from '../../lib/collab/provider'
 import ServiceClient from '../../lib/collab/stack-client'
 
-import { getShortNameFromClient, getParentFolderLink } from '../../lib/utils.js'
+import {
+  getShortNameFromClient,
+  getParentFolderLink,
+  getAppFullName
+} from '../../lib/utils.js'
 
 import { translate } from 'cozy-ui/react/I18n'
+
+function setPageTitle(appFullName, title) {
+  document.title =
+    title && title != '' ? `${appFullName} - ${title}` : appFullName
+}
 
 const Editor = translate()(
   withClient(function(props) {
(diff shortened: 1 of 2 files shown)


src/components/notes/editor.jsx
```diff
@@ -8,17 +8,27 @@ import EditorLoading from './editor-loading'
 import CollabProvider from '../../lib/collab/provider'
 import ServiceClient from '../../lib/collab/stack-client'
 
-import { getShortNameFromClient, getParentFolderLink } from '../../lib/utils.js'
+import {
+  getShortNameFromClient,
+  getParentFolderLink,
+  getAppFullName
+} from '../../lib/utils.js'
```
Crash--:
```suggestion
} from 'lib/utils.js'
```

src/lib/utils.js
```diff
@@ -4,6 +4,9 @@ import {
   generateWebLink,
   generateUniversalLink
 } from 'cozy-ui/transpiled/react/AppLinker'
+import { getDataset, getDataOrDefault } from './initFromDom'
```
Crash--:
```suggestion
import { getDataset, getDataOrDefault } from 'lib/initFromDom'
```

src/lib/utils.js
```diff
@@ -80,3 +83,13 @@ export function getFullLink(client, id = null) {
 export function getParentFolderLink(client, file) {
   return getFullLink(client, getParentFolderId(file))
 }
+
+export function getAppFullName() {
+  const dataset = getDataset()
```
Crash--: getDataSet() can be pretty extensive since we don't memoize it. You should use `client.getInstanceOptions()` and we should have a method for that https://github.com/cozy/cozy-notes/blob/master/src/components/app.jsx#L68-L69

edas: Well, if we cache it, it's not 2 ou 3 calls that will make a real difference

Crash--: Other argument; 

We plan to make getInstanceOption() works on mobile too (ie without having data injected by the stack). So it's better to rely on cozy-client for the futur, nop?

PR #31, opened 2019-12-19
feat: Fetch Notes Title

Crash--:
Since the stack doesn't give us a way to query the notes and that we want to fetch the note title without the stack's debounce we need to make a query for each note to get its title.

TODO : create NotesCollection on Cozy-Client


Crash--:
https://github.com/cozy/cozy-client/issues/601

PR #39, opened 2019-12-23
tests: Fix test command

Crash--:
We had an error when we tried to run an empty test. This error was the result of a not enough restrictive moduleNameMapper.

This is not the first time, we should pay attention to this.


y-lohse:
Nice work, that error message was completely misleading.

PR #45, opened 2020-01-02
fix: EditorBar issue on mobile

Crash--:
Since `Layout` adds a `before` with an height of 3rem (https://github.com/cozy/cozy-ui/blob/master/stylus/objects/layouts.styl#L105) on medium-screen to give to the bar the needed place and since we hide the cozy-bar, we had to remove this. I just used a -3rem marginTop


Crash--:
Des snapshots pourraient servir à l'avenir pour détecter ce genre de pb :)

PR #53, opened 2020-01-06
feat: Sync on leaving a note

edas:
With this change, we send a sync() request to the server
when leaving a note. This will allows the server to always
have files names and content up to date when we finish
to edit a note (instead of having to wait a few minutes).

Collateral: We now go back to the index with an anchor
and not by reloading the full page.

### DIFF
--- a/src/components/notes/editor.jsx
+++ b/src/components/notes/editor.jsx
@@ -26,6 +26,7 @@ import {
 } from 'lib/utils.js'
 
 import { translate } from 'cozy-ui/react/I18n'
+import useEventListener from 'cozy-ui/react/hooks/useEventListener'
 
 function setPageTitle(appFullName, title) {
   document.title =
@@ -103,6 +104,9 @@ const Editor = translate()(
             { version: doc.version, docId },
             serviceClient
           )
+          // The following object is defined in an Atlassian API.
+          // `provider` expects a Promise, even if we wouldn't
+          //  need it ourselves.
           return {
             useNativePlugin: true,
             provider: Promise.resolve(provider),
@@ -179,6 +183,27 @@ const Editor = translate()(
       },
       [loading, doc]
     )
+
+    // Be sure to save everything before leaving
+    // and force sync with the io.cozy.file
+    async function forceSync() {
+      if (doc) {
+        // wait for every event to finish
+        const provider = await collabProvider.provider
+        await provider.channel.ensureEmptyQueue()
+        // then force a server sync
+        await serviceClient.sync(docId)
+      }
+    }
+    useEffect(() => forceSync, [docId, doc])
+    // Sync on unload will probably be stopped by the browser,
+    // as most async code on unload, but let's try anyway
+    const emergencySync = useCallback(function() {
+      if (doc && docId) {
+        serviceClient.sync(docId) // force a server sync
+      }
+    }, [])
+    useEventListener(window, 'unload', emergencySync)
 
     // rendering
     if (loading) {
--- a/src/lib/collab/channel.js
+++ b/src/lib/collab/channel.js
@@ -5,6 +5,20 @@ const minimumBackoff = 128 // 128ms
 const maximumBackoff = 1000 * 60 * 5 // Max 5 minutes
 const failuresBeforeCatchup = 4
 
+const resolvablePromise = () => {
+  let resolveFunction
+  const promise = new Promise(resolve => {
+    resolveFunction = resolve
+  })
+  promise.resolved = false
+  promise.resolve = function() {
+    resolveFunction()
+    promise.resolved = true
+    return promise
+  }
+  return promise
+}
+
 export class Channel {
   constructor(config, serviceClient) {
     this.config = config
@@ -13,6 +27,7 @@ export class Channel {
     this.resetBackoff()
     this.initializeStepsQueue()
     this.isSending = false
+    this.emptyQueuePromise = resolvablePromise().resolve()
   }
 
   /**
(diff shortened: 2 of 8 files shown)


src/components/notes/editor.jsx
```diff
@@ -26,6 +26,7 @@ import {
 } from 'lib/utils.js'
 
 import { translate } from 'cozy-ui/react/I18n'
+import useEventListener from '../../hooks/useEventListener'
```
Crash--: alias path plz

src/hooks/useEventListener.js
```diff
@@ -0,0 +1,38 @@
...
+      element.addEventListener(eventName, eventListener)
+
+      // Remove event listener on cleanup
+      return () => {
+        element.removeEventListener(eventName, eventListener)
+      }
+    },
+    [eventName, element] // Re-run if eventName or element changes
+  )
+}
```
Crash--: Should we mutualize here? https://github.com/cozy/cozy-ui/blob/master/react/hooks/useEventListener.js this one has already a spec file you can improve ;)

src/lib/collab/channel.js
```diff
@@ -5,6 +5,20 @@ const minimumBackoff = 128 // 128ms
...
+    resolveFunction = resolve
+  })
+  promise.resolved = false
+  promise.resolve = function() {
+    resolveFunction()
+    promise.resolved = true
+    return promise
+  }
+  return promise
+}
```
Crash--: This should be extracted somewhere else with a js doc to explain the goal of this function

Crash--: We already have a cancelablePromise, we should add the resolvable there https://github.com/cozy/cozy-client/blob/1431dcbbe30af8b1b1eb731c065aee146d26caa4/packages/cozy-client/src/utils.js#L9

src/lib/collab/stack-client.js
```diff
@@ -114,6 +114,14 @@ export class ServiceClient {
     return this.stackClient.fetchJSON('POST', this.path(), doc)
   }
 
+  /**
+   * Force the note to be written to the io.cozy.files now
+   * @param {uuid} docId - note id
+   */
+  async sync(docId) {
+    await this.stackClient.fetchJSON('POST', this.path(docId, 'sync'))
```
Crash--: we should just return the promise, no?

src/components/notes/editor.jsx
```diff
@@ -180,6 +181,27 @@ const Editor = translate()(
       [loading, doc]
     )
 
+    // Be sure to save everything before leaving
+    // and force sync with the io.cozy.file
+    async function forceSync() {
+      if (doc) {
+        // wait for every event to finish
+        const provider = await collabProvider.provider
```
Crash--: Why are we returning : `Promise.resolve(provider)` ? If this task is async we should use a fetcher to tell so. `await collabProvider.fetchProvider()` is clearer. But I'm curious about the promise.resolve

src/lib/collab/stack-client.js
edas: I am unsure about the benefit, but it won't harm anyway, 👍

edas: What would you expect and for what usage ?

src/components/notes/editor.jsx
edas: `collabProvider`'s API defined by Atlassian. It expects a Promise at the `provider` key. Our own provider is initialized without needing any async code. We use `Promise.resolve()` to comply with Atlassian's API.

Crash--: OK thanks. Adding a comment around `Promise.resolve(provider)` can be useful I think :)

src/lib/collab/stack-client.js
Crash--: What is the benefit to await here? 

I think we'll do : 
`await client.sync()`

so 
`await await this.stackClient.fetchJSON('POST', this.path(docId, 'sync'))` brings nothing

and if we only do `client.sync()` it'll not wait the end of the call. So there is no benefits to await here, no?

src/lib/collab/channel.js
ptbrowne: What is the difference with Promise.resolve ?

src/lib/collab/channel.js
```diff
@@ -60,13 +75,23 @@ export class Channel {
     this.queuedStep = undefined
   }
 
+  /**
+   * Ensure all local steps are sent to the server
+   */
+  async ensureEmptyQueue() {
+    return this.waitForEmptyQueue
```
ptbrowne: ~~Are we sure we do not want to call the function here ?~~

I think this.waitForEmptyQueue should be renamed, it seems like it is a method when it is a promise. We should avoid using verbs for non function variables.

src/lib/collab/channel.js
```diff
@@ -60,13 +75,23 @@ export class Channel {
     this.queuedStep = undefined
   }
 
+  /**
+   * Ensure all local steps are sent to the server
```
ptbrowne:
```suggestion
   * Ensures all local steps are sent to the server
```

src/lib/collab/channel.js
edas: You can resolve later. 
It was just a local tool to avoid using two properties, one for the promise and one to externalize the resolution function.

edas: https://github.com/cozy/cozy-client/pull/610

But I am still unsure about the benefit of having it extracted

src/lib/collab/stack-client.js
edas: Either you `return this.stackClient.fetchJSON()` or you `await this.stackClient.fetchJSON()`.

Both will have the same async behaviour (if you wait `client.sync()`, it will wait the  `fetchJSON` from cozyClient). The difference is that the later does always return `undefined` and does not create a implicit API that someone may rely on. As the result of `fetchJSON` is useless to the caller, I would prefer not to return it.

src/lib/collab/channel.js
edas: -> emptyQueuePromise

src/lib/collab/stack-client.js
edas: (I don't see any blocker anyways, so if it's important for you having this with a return, I will add a return)

Crash--: It's not a blocker for me neither. Your explanation is largely convincing ;)

PR #59, opened 2020-01-08
feat: Force sync when leaving the title

edas:
Force a sync on the underlying io.cozy.file
when the user leave the title field.
This will force the io.cozy.file to update its
filename with the lastest title.

### DIFF
--- a/src/components/notes/editor.jsx
+++ b/src/components/notes/editor.jsx
@@ -198,11 +198,14 @@ const Editor = translate()(
     useEffect(() => forceSync, [docId, doc])
     // Sync on unload will probably be stopped by the browser,
     // as most async code on unload, but let's try anyway
-    const emergencySync = useCallback(function() {
-      if (doc && docId) {
-        serviceClient.sync(docId) // force a server sync
-      }
-    }, [])
+    const emergencySync = useCallback(
+      function() {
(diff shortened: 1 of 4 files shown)


src/components/notes/editor.jsx
```diff
@@ -198,11 +198,15 @@ const Editor = translate()(
...
     // Sync on unload will probably be stopped by the browser,
     // as most async code on unload, but let's try anyway
-    const emergencySync = useCallback(function() {
-      if (doc && docId) {
-        serviceClient.sync(docId) // force a server sync
-      }
-    }, [])
+    const emergencySync = useCallback(
+      function() {
+        console.log('emergencySunc run', doc, docId)
```
ptbrowne: console.log should be removed

edas: oops. this wasn't catched by the linter

ptbrowne: Weird that the linter did not catch this, this might hide something bigger ?

edas: Yep, it seems that `console` does not trigger the linter on cozy-script v2.

edas:
> Should this be tested ?

I have a major refactoring in WIP that will allow more things to be tested

PR #60, opened 2020-01-08
refactor: Simplify <Editor> with hooks

edas:
Tentative de refactorisation de &lt;Editor>

Objectif de limiter la complexité (je commençais à ne plus forcément voir toutes les interactions) et de permettre plus tard le test de la logique métier (qui me semblait plus que difficile à tester dans la forme monolithique)

### DIFF
--- a/src/components/notes/editor-view.jsx
+++ b/src/components/notes/editor-view.jsx
@@ -15,17 +15,19 @@ function updateTextareaHeight(target) {
   target.style.height = `${target.scrollHeight}px`
 }
 
+const nullCallback = () => {}
+
 function EditorView(props) {
   const {
     defaultValue,
     onTitleChange,
     onTitleBlur,
-    onContentChange,
     defaultTitle,
     title,
     collabProvider,
     leftComponent,
     rightComponent,
+    onContentChange,
     t
   } = props
 
@@ -55,7 +57,7 @@ function EditorView(props) {
       <section className="note-editor-container">
         <Editor
           collabEdit={collabProvider}
-          onChange={onContentChange}
+          onChange={onContentChange || nullCallback}
           defaultValue={defaultValue}
           {...editorConfig}
           appearance="full-page"
--- a/src/components/notes/editor.jsx
+++ b/src/components/notes/editor.jsx
@@ -1,10 +1,4 @@
-import React, {
-  useCallback,
-  useState,
-  useEffect,
-  useMemo,
-  useContext
-} from 'react'
+import React, { useEffect, useContext } from 'react'
 
 import { withClient } from 'cozy-client'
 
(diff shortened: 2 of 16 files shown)


Crash--:
OMG j'ai pas encore relu, mais un grand OUI pour l'effort 👏

src/hooks/useForceSync.js
```diff
@@ -0,0 +1,35 @@
...
+    function() {
+      if (doc && serviceClient) {
+        serviceClient.sync(noteId) // force a server sync
+      }
+    },
+    [doc, serviceClient]
+  )
+
+  return { forceSync, emergencySync }
+}
```
ptbrowne: Should forceSync and emergencySync be methods on the service ?

I think that there should be the least amount of code inside hooks, so that they can live outside React.

src/components/notes/editor.jsx
```diff
@@ -13,196 +7,58 @@ import EditorLoading from 'components/notes/editor-loading'
...
-  getParentFolderLink,
-  getAppFullName
-} from 'lib/utils.js'
+import useNote from 'hooks/useNote'
+import useServiceClient from 'hooks/useServiceClient'
+import useCollabProvider from 'hooks/useCollabProvider'
+import useTitleChanges from 'hooks/useTitleChanges'
+import useForceSync from 'hooks/useForceSync'
+import useReturnUrl from 'hooks/useReturnUrl'
+import useUser from 'hooks/useUser'
```
ptbrowne: Lot of hooks, I would be interested to know if they are really decoupled and usable in a standalone manner, each one from each one. At this point I am wondering if a class component would not be better than many decoupled hooks.

src/hooks/useForceSync.js
edas: En fait le fonctionnel est déjà dans `serviceClient.sync` et `channel.ensureEmptyQueue`. Ces deux méthodes gèrent l'interaction avec les états du composant (est ce que le doc est chargé, est-ce que les services sont prêts)

src/hooks/useForceSync.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+    },
+    [doc, collabProvider]
+  )
+
+  // Sync on unload will probably be stopped by the browser,
+  // as most async code on unload, but let's try anyway
+  const emergencySync = useCallback(
+    function() {
+      if (doc && serviceClient) {
```
ptbrowne: Ca serait plus clair si c'était `if (noteId && serviceClient)` non ? On se demande un peu pourquoi on if sur le doc alors qu'on l'utilise pas en suite.

src/components/notes/editor.jsx
edas: Most of them are pretty specific to this component will probably never be used in another context.

src/hooks/useForceSync.js
ptbrowne: Je trouve ça un peu perturbant d'avoir la méthode nommée emergencySync alors qu'en soit, elle ne fait rien en urgence, c'est le contexte appelant qui l'appelle en urgence.

src/components/notes/editor-view.jsx
```diff
@@ -37,6 +36,14 @@ function EditorView(props) {
...
   )
 
+  const onContentChange = useMemo(
+    () => {
+      const nullCallback = () => {}
+      return props.onContentChange || nullCallback
+    },
+    [props.onContentChange]
+  )
+
```
y-lohse: Why is this better than straight up using the prop?

src/hooks/useForceSync.js
```diff
@@ -0,0 +1,35 @@
+import { useCallback } from 'react'
+
+function useForceSync({ doc, serviceClient, collabProvider }) {
+  const noteId = doc && doc.file.id
+
+  // Be sure to save everything before leaving
+  // and force sync with the io.cozy.file
+  const forceSync = useCallback(
+    async function forceSync() {
```
y-lohse: Nit but out codebases usually have anonymous functions in this sort of cases.

src/hooks/useNote.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // reload if ever noteId changes
+  useEffect(
+    () => {
+      if (docId !== noteId) {
+        setLoading(true)
+        setTitle(undefined)
+        setDoc(undefined)
+        setDocId(noteId)
+      }
```
y-lohse: Does this hook make sense without the other `useEffect` in this file? Would it make sense to combine them into a single `useEffect` call?

src/hooks/useNote.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    [noteId, docId, setLoading, setTitle, setDoc, setDocId]
```
y-lohse: I've never seen the setters as a dependency, is it useful? I would assumre they never change during the course of the hooks llfecycle, no?

src/hooks/useNote.js
```diff
@@ -0,0 +1,52 @@
...
+          const doc = await serviceClient.getDoc(docId)
+          setTitle(doc.title || '')
+          setDoc(doc)
+        } catch (e) {
+          setTitle(false)
+          setDoc(false)
+          // eslint-disable-next-line no-console
+          console.warn(`Could not load note ${docId}`)
+        }
+        setLoading(false)
```
y-lohse: Should this be in a `finally` block?

src/hooks/useNote.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+      if (serviceClient) {
+        if (!doc || doc.file.id != docId) loadNote()
+      }
+    },
+    [docId, setLoading, serviceClient, setDoc, setTitle]
+  )
+
+  if (docId == noteId) return { loading, title, doc, setTitle }
+  else return { loading: true, title: undefined, doc: undefined, setTitle }
```
y-lohse: Aren't these values set though the first `useEffect` anyway?

src/hooks/useTitleChanges.js
```diff
@@ -0,0 +1,43 @@
+import { useMemo, useEffect, useCallback } from 'react'
+import { getAppFullName } from 'lib/utils.js'
+
+function useTitleChanges({ noteId, title, setTitle, serviceClient }) {
+  // Title change because of a local change from the editor
+  const onLocalTitleChange = useCallback(
+    serviceClient
+      ? e => {
+          const newTitle = e.target.value
+          const modifiedTitle = newTitle
```
y-lohse: We only need one of these 2 variables.

src/hooks/useTitleChanges.js
```diff
@@ -0,0 +1,43 @@
...
+          if (title != modifiedTitle) {
+            setTitle(modifiedTitle)
+            serviceClient.setTitle(noteId, modifiedTitle)
+          }
+        }
+      : () => {},
+    [noteId, title, setTitle, serviceClient]
+  )
+  // Title change because of a remote change from the stack
+  useMemo(
```
y-lohse: This should be a `useEffect`, right? `useMemo` is for memoized values, but this function doesn't return anything.

src/components/notes/editor.jsx
```diff
@@ -13,196 +7,58 @@ import EditorLoading from 'components/notes/editor-loading'
...
-    // fetch the actual note on load
-    useEffect(
-      () => {
-        loadNote(serviceClient, noteId, loading, setLoading, setDoc, setTitle)
-      },
-      [noteId]
-    )
+    const isPublic = useContext(IsPublicContext)
+    const { userName, userId } = useUser({
+      userName: props.userName,
```
y-lohse: Nit, but might as well destructure `userName ` from `props` at the start with the other props. It makes it explicit what props are being used.

src/components/notes/editor-view.jsx
edas: None, you're right. Let me change that.
Thank you

src/hooks/useForceSync.js
edas: Yep, that's a few times that I wanted to raise the subject.

A plain old function allows us to have a named function in our debug tools and call stack, with no additionnal shortcomming that I know of.

It also remove the need to add an explicit displayName for React components.

Did I miss something that should make us use anonymous functions ?

ptbrowne: > It also remove the need to add an explicit displayName for React components.

Only when they are wrapped in an HOC at assignment time, otherwise the displayName is correctly inferred from the variable name.

Apart from this, I have no problem personally with the use of named function expressions.

src/hooks/useNote.js
edas: From the doc "The identity of the setCount function is guaranteed to be stable so it’s safe to omit".

So no, in this case (straight `useState` without wrapping or custom modification), it's not needed.

I see it as a good practice to always list all dependencies anyways. It won't trigger any  problem and let us have a an habit that to not require us to check where the setter is from, nor forgive to add a setter in the list later if we change where it comes from.

https://stackoverflow.com/a/57180908

src/hooks/useNote.js
edas: The only difference I know of, is if the catch block is itself throwing an exception. If this happens here, I don't think we want to trigger an end of loading state.

src/hooks/useNote.js
edas: nope.

- You run the `useNote({noteId: 1})`, load the note 1, fill `title`, `doc` and `loading` for `docId=1`.
- Then you run `useNote({noteId: 2})` in a future render.  The effect is async and won't execute immediatly (it will be planned just after the render of the component calling `useNote`). You will then reach the end of `useNote` with noteId=2, docId=1, and the title/doc/loading values of docId=1, which is not  what the calling component has asked.

Once you have an effect, you should always think that your data may lag of one render, and add safeguards to avoid filling obsolete values

src/components/notes/editor.jsx
edas: I have that pattern two or three times in notes. I welcome suggestions on what to to.

Not destructuring allows me here to use `userName` everywhere whitout having one variable for the input and one for the output (and be sure that if I use `userName`, I'm sure it is a normalized value and not some raw thing that I shouldn't use)

src/hooks/useForceSync.js
edas: TIL, thank you

src/hooks/useNote.js
```diff
@@ -0,0 +1,46 @@
...
+
+      if (docId !== noteId) {
+        // reload if ever noteId changes
+        setLoading(true)
+        setTitle(undefined)
+        setDoc(undefined)
+        setDocId(noteId)
+      } else if (serviceClient) {
+        // load the note if needed
+        if (!doc || doc.file.id != docId) loadNote()
```
Crash--:
```suggestion
        if (!doc || doc.file.id !== docId) loadNote()
```

src/hooks/useNote.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+    },
+    [noteId, docId, setDocId, setLoading, serviceClient, setDoc, setTitle]
+  )
+
+  if (docId == noteId) return { loading, title, doc, setTitle }
```
Crash--:
```suggestion
  if (docId === noteId) return { loading, title, doc, setTitle }
```

src/lib/collab/channel.spec.js
```diff
@@ -5,8 +5,8 @@ jest.mock('prosemirror-collab', () => {
   return { getVersion: jest.fn(), sendableSteps: jest.fn() }
 })
 
-const docId = 'myDocId'
-const config = { docId }
+const noteId = 'myDocId'
```
Crash--:
```suggestion
const noteId = 'myNoteId'
```

src/lib/collab/stack-client.js
```diff
@@ -155,11 +159,11 @@ export class ServiceClient {
...
   /**
    * Join a doc - listen to realtime events for this doc
-   * @param {uui} docId
+   * @param {uui} noteId
    */
-  async join(docId) {
+  async join(noteId) {
     const onRealtimeCreated = function(doc) {
-      if (doc.id == docId) {
+      if (doc.id == noteId) {
```
Crash--:
```suggestion
      if (doc.id === noteId) {
```

src/hooks/useServiceClient.js
```diff
@@ -0,0 +1,13 @@
+import { useMemo } from 'react'
+import ServiceClient from 'lib/collab/stack-client'
+
+function useServiceClient({ userId, userName, cozyClient }) {
+  return useMemo(
+    () => {
+      return new ServiceClient({ userId, userName, cozyClient })
```
ptbrowne: I find ServiceClient very generic, it's hard to know what are its responsibilities from here.

src/hooks/useTitleChanges.js
```diff
@@ -0,0 +1,42 @@
...
+    () => {
+      serviceClient &&
+        serviceClient.onTitleUpdated(noteId, modifiedTitle => {
+          if (title != modifiedTitle) {
+            setTitle(modifiedTitle)
+          }
+        })
+    },
+    [title, setTitle, serviceClient]
+  )
```
ptbrowne: There are two different strategies here to deal with a potentially false serviceClient, inside the function or outside, could this be refactored so that only 1 techniques is used ?

src/hooks/useTitleChanges.js
```diff
@@ -0,0 +1,42 @@
...
+    [noteId, title, setTitle, serviceClient]
+  )
+  // Title change because of a remote change from the stack
+  useEffect(
+    () => {
+      serviceClient &&
+        serviceClient.onTitleUpdated(noteId, modifiedTitle => {
+          if (title != modifiedTitle) {
+            setTitle(modifiedTitle)
+          }
```
ptbrowne: This hook adds an onTitleUpdated callback but does not deal with the cleanup of the callback, is it normal ? It is a bit surprising.

src/hooks/useTitleChanges.js
ptbrowne: Could the code be refactored so that the hooks do not have to deal with the presence of serviceClient ? It seems cumbersome and error prone.

src/components/notes/editor.jsx
ptbrowne: What I mean was : all this hooks have to be used in combination and it seems that we gain nothing with having them in different files since they are very coupled in functionality. Could there be only one hook useEditor or useNote doing all these things ?

edas: I have mixed feelings about that. My initial purpose was to explode all logic in different files in order to avoid a unique very long file.

The Editor does not have really more than that. If we put everything in one unique hook, I'm afraid that we'll end un with exactly what we had before this PR.

src/hooks/useTitleChanges.js
edas: in serviceClient, we only have one callback per event, so each new `onTitleUpdated` replace the previous one.

src/hooks/useTitleChanges.js
edas: I've tried a few things, but with no success. I tried to first explode the logic to simplify Editor. We should be able to change the content of each hook in a future step

src/hooks/useServiceClient.js
edas: Ok about that, but it should probably be a separate issue.

-> https://github.com/cozy/cozy-notes/issues/64

src/components/notes/editor.jsx
ptbrowne: OK.

> My initial purpose was to explode all logic in different files in order to avoid a unique very long file.

I think the long file was not the problematic thing before, it was the super long render method. 

> If we put everything in one unique hook, I'm afraid that we'll end un with exactly what we had before this PR.

Extracting each hook and keeping them in the same file would have been OK for me. I personnally prefer when highly cohesive methods are in the same file, I find it easier to refactor and see the overall picture. However when things are not cohesive, sure, separate files are neded.

Here, Editor has to have knowledge of how to assemble all the little hooks, I have trouble saying why, but it troubles me that the orchestration of all those hooks is done in the render.

src/components/notes/editor.jsx
y-lohse: Aaah I see, that's fair enough. I don't have a better idea on how to handle this at the moment, with the explanation I also think not destructuring is better in this case.

ptbrowne: One solution would be to use `const { userName: propUserName } = props` ?

src/hooks/useNote.js
y-lohse: Mmh ok. I think the `noteId`/`docId` is a bit confusing and there might be a better wya to handle this, but it wasn't the scope of the PR and it's already easier to understand things with asingle hook, so LGTM 👍

PR #63, opened 2020-01-09
Feat/sharing label

edas:
Show the labelonsharing button for views > 1024px
We strip it on smaller views to leave space for the prosemirror toolbar.

Two things for which comments are welcome : 

* I sharing.styl I use a css rule to hide the label of the sharing button. This rule have to know how we render the button internals. I plan to add wrapper around the label in cozy-sharing, and a props allowing to inject a classname for this wrapper. Would this be ok for you?

* In index.css line 102-107 we have a useless block. Removing it would make the following block ignored. This is usually because of a syntax error above, but I fail to see where. 
This dirty hack was already present before this change, from the proof of concept, but I would welcome suggestions to remove it now.

### DIFF
--- a/src/components/notes/sharing.jsx
+++ b/src/components/notes/sharing.jsx
@@ -1,70 +1,32 @@
-import React, { useCallback, useState, useMemo, useEffect } from 'react'
+import React, { useCallback, useState } from 'react'
 import SharingProvider, { ShareButton, ShareModal } from 'cozy-sharing'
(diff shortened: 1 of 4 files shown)


y-lohse:
>  I plan to add wrapper around the label in cozy-sharing, and a props allowing to inject a classname for this wrapper. Would this be ok for you?

Not sure about the wrapper, but a prop to inject classnames is definitely welcome!

PR #68, opened 2020-01-16
feat: Title first title and then app name

Crash--:
If we have small tabs it's easier to have access to the note title first ;)


ptbrowne:
I am bit concerned that no tests have to be changed relative to this change though ;)

PR #71, opened 2020-01-20
Adds a confirmation on window exit

edas:
If there are changes not saved, this PR will adds a popup asking for confirmation before leaving the window.


src/hooks/useCollaborationState.js
```diff
@@ -0,0 +1,92 @@
...
+    if (dirtySince.current) {
+      dirtySince.current = null
+      setChanges(iterations++)
+    }
+  }
+
+  // Sets dirtySince when the queue fills up
+  function onEnqueueSteps() {
+    if (!dirtySince.current) {
+      if (!dirtySince.current) dirtySince.current = new Date()
```
y-lohse: Same condition twice here

src/hooks/useCollaborationState.js
```diff
@@ -0,0 +1,92 @@
...
+  const dirtySince = useRef(null)
+  const lastSave = useRef(null)
+  // the state is only here to trigger a rerender when needed
+  const [, setChanges] = useState(iterations)
+
+  // init
+  if (channel) {
+    if (channel.isDirty()) {
+      // always mark as dirty if the channel is dirty
+      if (!dirtySince.current) dirtySince.current = new Date()
```
y-lohse: Can `dirtySince.current` even be set at this stage?

src/hooks/useCollaborationState.spec.js
```diff
@@ -0,0 +1,338 @@
...
+        const { result } = await initAndQueue()
+        expect(result.current.isDirty).toBeTruthy()
+        await resolvePromise()
+      })
+
+      it('should have a new dirtySince date', async () => {
+        const { result, previous } = await initAndQueue()
+        expect(result.current.dirtySince).not.toBe(previous.dirtySince)
+        await resolvePromise()
+      })
```
y-lohse: 👍 nice comprehensive test suite

src/hooks/useCollaborationState.js
edas: Vu et corrigé, merci

src/hooks/useCollaborationState.js
edas: Most probably, you always start with a non dirty channel. I prefer to be defensive a prepare the case where the channel is dirty the first time I am executing this hook.

However, this specific line may well be executed even after initialization. If proseMirror enqueues a step, we'll fire an event. This event has to be catched. The event listener will then change de ref value and change the state to call a rerender.

All this is async. We may well have a rerender of the same component at any time, before all this, after all this, or between any step in the chain. So yes, we could have a dirty channel but still not have changed the ref value

package.json
```diff
@@ -50,20 +52,23 @@
     "@atlaskit/media-core": "^31.0.1",
     "@atlaskit/media-editor": "^37.0.3",
     "@atlaskit/smart-card": "^12.6.2",
+    "@material-ui/core": "^4.8.3",
```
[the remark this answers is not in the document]
edas: fixed

src/hooks/useCollabProvider.js
```diff
@@ -9,18 +9,21 @@ function useCollabProvider({ docVersion, noteId, serviceClient }) {
...
+          // The following `collabProviderPlugin` object is defined
+          // in an Atlassian API. The attribute `provider` is expected
+          // to be a Promise, even if we don't need one ourselves here.
+          collabProviderPlugin: {
+            useNativePlugin: true,
+            provider: Promise.resolve(provider),
+            inviteToEditHandler: () => undefined,
+            isInviteToEditButtonSelected: false,
+            userId: serviceClient.getSessionId()
+          }
```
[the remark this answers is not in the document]
edas: This part didn't really changed. I just return both the provider and the plugin where we returned only the plugin before the change.

PR #77, opened 2020-01-27
feat: Allows to turn the editor read-only

edas:
<Editor> and <EditorView> now have a `readOnly` attributes. 

If readOnly: 

- The toolbar will not be greyed out
- The reader cannot type into the editor or into the title
- The reader will still see updates coming from the server

### DIFF
--- a/src/components/notes/editor-view.jsx
+++ b/src/components/notes/editor-view.jsx
@@ -1,10 +1,10 @@
-import React, { useCallback, useRef, useEffect } from 'react'
+import React, { useCallback, useRef, useEffect, useMemo } from 'react'
 
 import { Editor, WithEditorActions } from '@atlaskit/editor-core'
 
 import { MainTitle } from 'cozy-ui/react/Text'
 import Textarea from 'cozy-ui/react/Textarea'
-import { translate } from 'cozy-ui/react/I18n'
+import { useI18n } from 'cozy-ui/react/I18n'
 import useEventListener from 'cozy-ui/react/hooks/useEventListener'
 
 import editorConfig from 'components/notes/editor_config'
@@ -12,7 +12,7 @@ import HeaderMenu from 'components/header_menu'
 import styles from 'components/notes/editor-view.styl'
 
 function updateTextareaHeight(target) {
-  target.style.height = `${target.scrollHeight}px`
+  if (target) target.style.height = `${target.scrollHeight}px`
 }
 
 const nullCallback = () => {}
@@ -28,8 +28,9 @@ function EditorView(props) {
     leftComponent,
     rightComponent,
     onContentChange,
-    t
+    readOnly
   } = props
+  const { t } = useI18n()
 
   const titleEl = useRef(null)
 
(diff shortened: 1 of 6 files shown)


ptbrowne:
Tests please ? :)

edas:
Celui là je veux bien de l'aide pour trouver un test pertinent

ptbrowne:
I think puting readOnly on EditorView and checking that onChange is null on Textarea would be ok ?

Nit: the conditions are not very readable IMHO. AMHA `onChange={readOnly ? nullCallback : onChange}` is much more readable than the alternativ with a negated Boolean : `{(!readOnly && onContentChange) || nullCallback}`.

Nit2: And the collabEdit prop could be renamed collabProvider ?)

src/components/notes/editor-view.jsx
```diff
@@ -42,6 +43,23 @@ function EditorView(props) {
     [onTitleChange]
   )
 
+  useMemo(() => collabProvider && collabProvider.setReadOnly(readOnly), [
+    readOnly,
+    collabProvider
+  ])
```
y-lohse: I find it odd to use `useMemo` but to discard the memoized value. I guess the goal is to avoid calling `collabProvider.setReadOnly(readOnly)` when `readOnly` hasn't changed?

Crash--:
> The reader won't see the changes to the doc in realtime (he will have to leave and load again the note to see the updates) 

On pert tout l'intérêt de note avec cette solution je trouve :/

src/components/notes/editor-view.jsx
```diff
@@ -1,18 +1,18 @@
...
+import { useI18n } from 'cozy-ui/react/I18n'
 import useEventListener from 'cozy-ui/react/hooks/useEventListener'
 
 import editorConfig from 'components/notes/editor_config'
 import HeaderMenu from 'components/header_menu'
 import styles from 'components/notes/editor-view.styl'
 
 function updateTextareaHeight(target) {
-  target.style.height = `${target.scrollHeight}px`
+  if (target) target.style.height = `${target.scrollHeight}px`
```
Crash--: does that means that `titleEl.current` can be null or undefined? how can it happen ?

edas: C'est la seule modif vraiment honteuse : Je n'ai pas réussi à faire fonctionner les attributs ref lors du mount par Enzime, et cette condition est là pour compenser la ref DOM qui n'est pas là dans ce cas.

edas:
> > The reader won't see the changes to the doc in realtime (he will have to leave and load again the note to see the updates)
> 
> On pert tout l'intérêt de note avec cette solution je trouve :/

Oui. J'ai poussé cette nuit une version qui garde le collaboratif. C'est celle là que tu as du passé en revue ce matin.

src/components/notes/editor-view.jsx
edas: yep

edas:
> Can `readOnly` be changed ? 

Yep, even if we probably won't change this attributes during a session, the editor will accept the change and react accordingly

> Will Editor react correctly to a change of collabEdit ?

I don't know. I suppose not. These are Atlaskit internals

src/components/notes/editor-view.jsx
ptbrowne: > Rappelez-vous que la fonction fournie à useMemo s’exécute pendant le rendu. N’y faites rien que vous ne feriez pas normalement pendant un rendu. Par exemple, les effets de bord doivent passer par useEffect, et non useMemo.

At first I was going to comment that useEffect could be more correct then I re-read the docs and found that the useMemo creator function was executed during render, so I guess this is normal and wanted to useMemo. I think it is not very clear though, can you add a comment ?

PR #79, opened 2020-01-27
feat: Adds a saving indicator

edas:
In large view, this changes adds a saving indicator
in the bottom right. It allows the user to know if a
save is in process or when was the last successful save.


src/components/notes/saving-indicator.styl
```diff
@@ -0,0 +1,9 @@
+.saving-indicator
+    position absolute
+    bottom 1rem
+    right 1rem
+    padding 0.25rem 0.75rem
+    color var(--cool-grey, #95999d)
+    background-color: #FFFFFFAA
+    @media (max-width: 1023px)
```
Crash--: you can use +medium-screen() I think https://github.com/cozy/cozy-ui/blob/master/stylus/settings/breakpoints.styl#L13

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,83 @@
...
+// constants
+const sec = 1000
+const min = 60 * sec
+const hour = 60 * min
+const day = 24 * hour
+
+// differents steps for message formating
+// [age, key, unit, coef] :
+// - `age`: the first list where the age of last save
+// if younger than this `age` will be used
```
drazik: I don't understand what you mean here. What is "the first list"? Do you mean that for a given last save time, the first item for which `age < now - last save time` is true will be used?

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,83 @@
...
+]
+
+/**
+ * @typedef {object} SavingIndicatorProps
+ * @property {Date} dirtySince - our copy was not in sync since this date
+ * @property {Date} lastSave - last time we managed to save something to the server
+ */
+
+/**
+ * Display  a saving indicator in the bottom right
```
drazik:
```suggestion
 * Display a saving indicator in the bottom right
```

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,83 @@
...
+  const { dirtySince, lastSave, t } = props
+  const now = new Date()
+
+  const [, setChange] = useState(0)
+  const rerender = () => {
+    setChange(val => val + 1)
+  }
+  const timeout = useRef(undefined)
+
+  if (timeout.current) window.clearTimeout(timeout.current)
```
drazik: Isn't it some kind of effect? It looks like it should be in a `useEffect`, no?

edas: These won't add any side effect in DOM, nor do need to wait until the component is rendered or attached in the DOM. AFAIK `useEffet` is useless here.

src/components/notes/saving-indicator.jsx
edas: I rephrased

drazik: Thanks

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+  [10 * sec, 'saved_just_now', sec, 1],
+  [min, 'saved_secs_ago', sec, 10],
+  [2 * min, 'saved_min_ago', min, 1],
+  [5 * min, 'saved_mins_ago', min, 1],
+  [20 * min, 'saved_mins_ago', min, 5],
+  [hour, 'saved_mins_ago', min, 10],
+  [2 * hour, 'saved_hour_ago', hour, 1],
+  [day, 'saved_hours_ago', hour, 1],
+  [2 * day, 'saved_day_ago', day, 1],
+  [9999 * day, 'saved_day_ago', day, 1]
```
y-lohse:
```suggestion
  [Infinity, 'saved_days_ago', day, 1]
```

2 suggestions here:

- A thousand days seems enough but might as well use `Infinity`, that's what it's there for
- You probably meant `saved_days_ago` with an S

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+  if (timeout.current) window.clearTimeout(timeout.current)
+
+  let message
+  if (!dirtySince) {
+    const diff = now - (lastSave || now)
+    const [, key, unit, coef] = steps.find(el => diff < el[0])
+    message = t(`Notes.SavingIndicator.${key}`, {
+      time: Math.floor(diff / unit / coef) * coef
+    })
+    window.setTimeout(rerender, unit * coef)
```
y-lohse: Should this be `timeout.current = window.setTimeout` or am I missing something?

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+  } else {
+    if (dirtySince < now - 30000) {
+      message = (
+        <>
+          {t('Notes.SavingIndicator.saving_impossible')}
+          <br />
+          {t('Notes.SavingIndicator.new_tentative')}
```
y-lohse: `tentative` means `provisoire`, `new_attempt` would be a better key imo.

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+  let message
+  if (!dirtySince) {
+    const diff = now - (lastSave || now)
+    const [, key, unit, coef] = steps.find(el => diff < el[0])
+    message = t(`Notes.SavingIndicator.${key}`, {
+      time: Math.floor(diff / unit / coef) * coef
+    })
+    window.setTimeout(rerender, unit * coef)
+  } else {
+    if (dirtySince < now - 30000) {
```
y-lohse: Could we have a constant for `3000` as well?

src/components/notes/saving-indicator.styl
```diff
@@ -0,0 +1,12 @@
+@require 'settings/breakpoints.styl'
+
+.saving-indicator
+    position absolute
+    bottom 1rem
+    right 1rem
+    padding 0.25rem 0.75rem
+    color var(--cool-grey, #95999d)
+    background-color: #FFFFFFAA
```
y-lohse:
```suggestion
    background-color: #FFFFFF
```

src/locales/en.json
```diff
@@ -22,6 +22,19 @@
     "Empty": {
       "welcome": "Welcome on Cozy Notes",
       "after_created": "Once your Note is created, you can find it in your "
+    },
+    "SavingIndicator": {
+      "saved_just_now": "Saved just now.",
+      "saved_secs_ago": "Saved %{time} sec ago.",
+      "saved_min_ago": "Saved %{time} min ago.",
+      "saved_mins_ago": "Saved %{time} min ago.",
```
y-lohse:
```suggestion
      "saved_mins_ago": "Saved %{time} mins ago.",
```

src/locales/en.json
```diff
@@ -22,6 +22,19 @@
...
+      "saved_hour_ago": "Saved %{time} hour ago.",
+      "saved_hours_ago": "Saved %{time} hours ago.",
+      "saved_day_ago": "Saved %{time} day ago.",
+      "saved_days_ago": "Saved %{time} days ago.",
+      "saving": "Saving...",
```
y-lohse:
```suggestion
      "saving": "Saving…",
```
(sorry, typography nerding, feel free to ignore these 😅)

src/locales/en.json
```diff
@@ -22,6 +22,19 @@
...
+      "saving_impossible": "Impossible to save.",
+      "new_tentative": "New tentative..."
```
y-lohse:
```suggestion
      "new_tentative": "New tentative…"
```

src/locales/fr.json
```diff
@@ -22,6 +22,19 @@
...
+    "SavingIndicator": {
+      "saved_just_now": "Enregistré à l'instant.",
+      "saved_secs_ago": "Enregistré il y a %{time} sec.",
+      "saved_min_ago": "Enregistré il y a %{time} min.",
+      "saved_mins_ago": "Enregistré il y a %{time} min.",
+      "saved_hour_ago": "Enregistré il y a %{time} heure.",
+      "saved_hours_ago": "Enregistré il y a %{time} heures.",
+      "saved_day_ago": "Enregistré il y a %{time} jour.",
+      "saved_days_ago": "Enregistré il y a %{time} jours.",
+      "saving": "Enregistrement en cours...",
```
y-lohse:
```suggestion
      "saving": "Enregistrement en cours…",
```

src/locales/fr.json
```diff
@@ -22,6 +22,19 @@
...
+      "saving_impossible": "Enregistrement impossible.",
+      "new_tentative": "Nouvelle tentative..."
```
y-lohse:
```suggestion
      "new_tentative": "Nouvelle tentative…"
```

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+// - `coef` is  the granularity in which we will send
+// the time (`min, 5` is "by 5 minutes slices")
+const steps = [
+  [10 * sec, 'saved_just_now', sec, 1],
+  [min, 'saved_secs_ago', sec, 10],
+  [2 * min, 'saved_min_ago', min, 1],
+  [5 * min, 'saved_mins_ago', min, 1],
+  [20 * min, 'saved_mins_ago', min, 5],
+  [hour, 'saved_mins_ago', min, 10],
+  [2 * hour, 'saved_hour_ago', hour, 1],
```
Crash--:
```suggestion
  [2 * hour, 'saved_hours_ago', hour, 1],
```
?

src/components/notes/saving-indicator.styl
```diff
@@ -0,0 +1,12 @@
+@require 'settings/breakpoints.styl'
+
+.saving-indicator
+    position absolute
+    bottom 1rem
+    right 1rem
+    padding 0.25rem 0.75rem
+    color var(--cool-grey, #95999d)
```
Crash--: isn't coolGrey ? Or do we expose both?

src/components/notes/saving-indicator.styl
edas: Nope, I really want the alpha transparency

src/locales/en.json
edas: nope, AFAIK, we usually not put a "s" for "40 min" in english

src/locales/en.json
edas: I did hesitate, but these are more loading signs that suspension points, and people may not recognize them as such  with suspension points.

src/locales/en.json
edas: (same here)

src/components/notes/saving-indicator.jsx
edas: nope.  The condition is "when time < 2 * hours". Which means it will be 1 hour (without "s")

src/components/notes/saving-indicator.styl
y-lohse: Almost ashamed to say this, but TIL. Always thought the only way to get transparency was to use `rgba`!

src/locales/en.json
y-lohse: Yeeah fair enough 😀works for me

src/locales/en.json
y-lohse: Interesting, [it seems a bit less strict than that](https://english.stackexchange.com/questions/125110/abbreviation-of-units-of-measure-hours-vs-minutes/125111) but `min` for `minutes` seems widely accepted!

Crash--:
Can we have a test for this feature?

src/components/notes/saving-indicator.jsx
ptbrowne: There is a side effect on window. AMHA the fact that setTimeout is not pure is sufficient to have this in useEffect.

AMHA we should extract all unpure logic outside the component and think of a way to make the component render pure.

Here, I think the intention of setTimeout is not very clear:

Could we have something like 

```
usePeriodicRender(shouldRerender, checkPeriod)
```

encapsulating the window.setTimeout/clearTimeout ?

---

Also I think it would be better to clear the timeout on unmount. I do not think it is done here.

src/components/notes/saving-indicator.jsx
```diff
@@ -0,0 +1,84 @@
...
+ * @property {Date} lastSave - last time we managed to save something to the server
+ */
+
+/**
+ * Displays a saving indicator in the bottom right
+ *
+ * @param {SavingIndicatorProps} props
+ */
+const SavingIndicator = translate()(props => {
+  const { dirtySince, lastSave, t } = props
```
ptbrowne: You can use useI18n instead of translate()

src/components/notes/saving-indicator.jsx
edas: Hum… I like that idea :+1:

edas:
La PR a beaucoup évolué suite aux retours de @ptbrowne 

J'ai retiré le useCollaborationState() pour me reposer directement sur la source et éviter des références. J'ai fait du coup pas mal grossir le provider mais ça devrait être plus compréhensible.

Je pousse mais ça reste en WIP le temps que j'arrive à faire ce que je veux sur le test du composant lui-même.

edas:
Tests ok

PR #84, opened 2020-01-31
Set the editor as read-only for a share where we don't have write permissions

edas:
Set the editor as read-only for a share where we don't have write permissions

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -33,21 +33,32 @@ const PrivateContext = () => (
 
 const PublicContext = withClient(({ client }) => {
   const [sharedDocumentId, setSharedDocumentId] = useState(null)
+  const [readOnly, setReadOnly] = useState(false)
   const returnUrl = useMemo(() => getReturnUrl(), [])
 
-  useEffect(() => {
-    const fetchData = async () => {
-      try {
-        const id = await getSharedDocument(client)
-        setSharedDocumentId(id)
-      } catch {
-        setSharedDocumentId(false)
+  useEffect(
+    () => {
+      const fetchData = async () => {
+        try {
+          const { id, readOnly } = await getSharedDocument(client)
+          setReadOnly(readOnly)
+          setSharedDocumentId(id)
+        } catch {
+          setSharedDocumentId(false)
+        }
       }
-    }
-    fetchData()
-  }, [])
+      fetchData()
+    },
+    [client]
+  )
   if (sharedDocumentId) {
-    return <Editor noteId={sharedDocumentId} returnUrl={returnUrl || false} />
+    return (
+      <Editor
+        readOnly={readOnly}
+        noteId={sharedDocumentId}
+        returnUrl={returnUrl || false}
+      />
+    )
   } else if (sharedDocumentId !== null) {
     return <Unshared />
   } else {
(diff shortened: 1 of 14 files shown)


edas:
J'ai poussé un commit sur les dépendances de hooks pour valider le linter mais les deux commits sont séparés si ça facilite la revue

src/components/app.jsx
```diff
@@ -33,21 +33,32 @@ const PrivateContext = () => (
 
 const PublicContext = withClient(({ client }) => {
   const [sharedDocumentId, setSharedDocumentId] = useState(null)
+  const [readOnly, setReadOnly] = useState(false)
```
ptbrowne: This means that for some time dependning on the network, the document can be read only when the user does not have the right permissions. What happens in this case if there are modifications ? Is this handled ?

src/lib/utils.spec.js
```diff
@@ -0,0 +1,67 @@
+import { getSharedDocument } from './utils'
+
+describe('getSharedDocument', () => {
+  function setupClient(verbs = []) {
+    return {
+      collection: () => ({
+        getOwnPermissions: () => ({
```
ptbrowne: This should be an async method right ?

yarn.lock
```diff
@@ -3123,19 +3123,11 @@ anymatch@^2.0.0:
     micromatch "^3.1.4"
     normalize-path "^2.1.1"
 
-aproba@^1.0.3, aproba@^1.1.1:
+aproba@^1.1.1:
```
ptbrowne: Why are there these changes without package.json changes ?

src/components/app.jsx
edas: The changes will be refused by the server anyways and will be erased when the readOnly attribute will kick in. There should be a really bad timing for the editor to be ready before the readOnly being set but you're right, I'll chang the order  of a few lines below  to make sure it will never happen

edas: > What happens in this case if there are modifications ? Is this handled ?

If the editor is set readOnly after its first use, any modifications will be refused by the server anyways and will stay pending in the local editor.
When the editor goes into readonly, all these local changes will be deleted to only leave the server version.

> This means that for some time dependning on the network, the document can be read only when the user does not have the right permissions

Nope, the setReadOnly is always called before the setSharedDocumentId so we never initialize the editor prior to setting to readonly attribute.

yarn.lock
edas: IDK, removed

src/lib/utils.js
```diff
@@ -28,16 +28,40 @@ export const generateReturnUrlToNotesIndex = doc => {
...
     .getOwnPermissions()
-
   const permissions = permissionsData.attributes.permissions
   // permissions contains several named keys, but the one to use depends on the situation. Using the first one is what we want in all known cases.
-  const sharedDocumentId = get(Object.values(permissions), '0.values.0')
-
-  return sharedDocumentId
+  const perm = get(Object.values(permissions), '0')
+  const sharedDocumentId = perm && get(perm, 'values.0')
+  const readOnly = perm && isPermissionReadOnly(perm)
```
Crash--: All these "magicals" stuff should be done in cozy-client. (get perm and doc_id). Nope?

the check for the readOnly should also be mutualized.

edas: I have more on this in a future PR. Let me passe those two then we'll be able to see what could be put  in common. The feature as it is in this PR is still an ugly shortcut.

Crash--: can we have an issue or something to keep that in mind?

PR #94, opened 2020-02-10
fix: Goes back to root dir

edas:
When the note is  present in the root dir, the link going back if invalid and send to an error page.
This change fixes that.

Note: I do not know nor understand why we were stripping "-"  in the folder ids. It's probably better to wait for @Crash-- approval before merging this PR


Crash--:
I don't know neither why we are doing that. I think it was already here when I started to write the code. 

After a quick search in our code base, it seems that we don't do that so I'm pretty confident with this change ;)

PR #98, opened 2020-02-12
fix: Go back to previous folderId

edas:
We moved a function to cozy-clients but I forgot to update a component using it


edas:
> having a small test even if it just a mount / snapshot can save us for this kind of mistake

:+1: added now

PR #99, opened 2020-02-13
fix: Show the last save date

edas:
We were not using the correct date in the saving indicator. This change fixes this.


src/lib/collab/provider.js
```diff
@@ -395,6 +395,7 @@ export class CollabProvider {
    */
   onLocalSave() {
     this.lastLocalSave = new Date()
+    console.log('onLocalSave', this.lastLocalSave)
```
Crash--: We should remove this log right?

edas: yep, this wasa mistake, see https://github.com/cozy/cozy-notes/pull/100

PR #100, opened 2020-02-13
fix: Remove console.logs

edas:


ptbrowne:
How did they get past the linter ?

edas:
@ptbrowne This not the first time. Ihave to look into it tomorow.

Also, I can't understand why the tests on this PR are failing in Travis

nono:
@ptbrowne we don't have a rule in the linter to catch the `console.log`. Eslint is configured with 3 presets, `eslint:recommended`, `eslint-config-prettier`, and `react/recommended`, and none of those presets include the `no-console` rule of `eslint`.

nono:
And the `no-console` rule was removed from `eslint:recommended` with the version 6: https://github.com/eslint/eslint/blob/24ca088fdc901feef8f10b050414fbde64b55c7d/docs/user-guide/migrating-to-6.0.0.md#eslint-recommended-changes

ptbrowne:
Indeed, thanks for checking !

y-lohse:
> Also, I can't understand why the tests on this PR are failing in Travis

The problem seems to come from [these 2 lines](https://github.com/cozy/cozy-notes/blob/master/src/lib/collab/provider.spec.js#L500-L501). Since the `wait` is after the `emit` it is possible that the `emit` finished in the same millisecond as `before`, which is what seems to happen during the tests.

One fix could be to swap the 2 lines, waiting 50ms before doing the `emit`. Or we could replace `expect(before).not.toEqual(afterSync)` with `expect(before).not.toBe(afterSync)` and get rid of the 50ms wait. I think the latter is probably a better solution, what do you think?

y-lohse:
(Note that the same problem could appear in the remote sync test as well)

PR #101, opened 2020-02-13
feat: Allows a user provided username

edas:
In collaboration mode, we show initials attached to each cursors. We also plan to display avatars in the future.

This PR allows a cozy to send the editor name when building a link, and have the correct name first letter.

It also remove old placeholder data (email and avatar) which are not displayed today but were throwing CSP errors in the console.

### DIFF
--- a/src/hooks/useUser.js
+++ b/src/hooks/useUser.js
@@ -1,10 +1,13 @@
 import { useMemo } from 'react'
 
-import { getShortNameFromClient } from 'lib/utils.js'
+import { getShortNameFromClient, getUserNameFromUrl } from 'lib/utils.js'
 
 function useUser({ userName: providedUserName, cozyClient }) {
   const userName = useMemo(
-    () => providedUserName || getShortNameFromClient(cozyClient),
+    () =>
+      providedUserName ||
+      getUserNameFromUrl() ||
+      getShortNameFromClient(cozyClient),
     [cozyClient, providedUserName]
   )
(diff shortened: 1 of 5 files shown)


ptbrowne:
What is the goal, I have trouble understanding in which context this is used.

Crash--:
Does that mean that now we display a badge with the Username passed in the URL?

edas:
In collaboration mode, we show initials attached to each cursors. We also plan to display avatars in the future.

This PR allows a cozy to send the editor name when building a link, and have the correct name first letter.

edas:
Display of avatars is not included in this PR

src/hooks/useUser.js
```diff
@@ -1,10 +1,13 @@
...
 
-import { getShortNameFromClient } from 'lib/utils.js'
+import { getShortNameFromClient, getUserNameFromUrl } from 'lib/utils.js'
 
 function useUser({ userName: providedUserName, cozyClient }) {
   const userName = useMemo(
-    () => providedUserName || getShortNameFromClient(cozyClient),
+    () =>
+      providedUserName ||
+      getUserNameFromUrl() ||
```
Crash--: We're passing thins value down to prosemirror collab service by emitting a joined event. Are we sure that the consumer of this value are escaping the value?

edas: Yep, the mock of the collab service is also using names with a space

PR #102, opened 2020-02-13
feat: Change saving indicator wordings

edas:
These wordings are for the saving indicator in the bottom right in large views (not displayed yet in small views). Thenew wordings are the one requested by @joel-costa for a better UX

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -154,6 +154,67 @@ export function createNoteDocument(client) {
   })
 }
 
+// constants
+const sec = 1000
+const min = 60 * sec
+const hour = 60 * min
+const day = 24 * hour
+const month = 31 * day
+const year = 365 * day
+
+// differents steps for message formating
+// [age, key, unit, coef] :
+// - `age`: the line applies only if the last saved
+// was done less than this `age` ago
+// (only the first match will apply)
+// - `key` is the traduction key for the message
+// - `unit` is the unit in which we will send the time
+// to the translated message
+// - `coef` is  the granularity in which we will send
+// the time (`min, 5` is "by 5 minutes slices")
+const defaultSteps = [
+  [10 * sec, 'just_now', sec, 1],
+  [1 * min, 'secs_ago', sec, 10],
+  [2 * min, 'min_ago', min, 1],
+  [5 * min, 'mins_ago', min, 1],
+  [20 * min, 'mins_ago', min, 5],
+  [1 * hour, 'mins_ago', min, 10],
+  [2 * hour, 'hour_ago', hour, 1],
+  [1 * day, 'hours_ago', hour, 1],
+  [2 * day, 'day_ago', day, 1],
+  [1 * month, 'days_ago', day, 1],
+  [2 * month, 'month_ago', month, 1],
+  [1 * year, 'monthes_ago', month, 1],
+  [2 * year, 'year_ago', year, 1],
+  [Infinity, 'years_ago', year, 1]
+]
+
+/**
+ * @typedef {object} RelativeAge
+ * @property {string} key - translation (sub) key
+ * @property {unit} unit - unit (minute, second, hour…) in which the age will be displayed
+ * @property {integer} coef - in how much values of this unit do we count (like every 10 minutes)
+ * @property {integer} time - age to be inserted in the translation message
+ * @property {integer} interval - `unit * coef`: compute again in that timeframe (in milliseconds)
+ */
+/**
+ * Gets raw data to display a relative age in human words
+ * @param {integer} age - age in milliseconds
+ * @param {array} steps - array of [max_age, key, unit, coef]
+ * @returns {RelativeAge}
+ */
+export function relativeAge(age, steps) {
+  const step = (steps || defaultSteps).find(el => age < el[0])
+  const [, key, unit, coef] = step
+  return {
+    key,
+    unit,
+    coef,
+    time: Math.floor(age / unit / coef) * coef,
+    interval: unit * coef
+  }
+}
+
 /**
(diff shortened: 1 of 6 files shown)


src/locales/fr.json
```diff
@@ -31,16 +31,16 @@
...
-      "saved_day_ago": "Enregistré il y a %{time} jour.",
-      "saved_days_ago": "Enregistré il y a %{time} jours.",
+      "saved_just_now": "Dernière mise à jour à l'instant.",
+      "saved_secs_ago": "Dernière mise à jour il y a %{time} sec",
+      "saved_min_ago": "Dernière mise à jour il y a %{time} min",
+      "saved_mins_ago": "Dernière mise à jour il y a %{time} min",
+      "saved_hour_ago": "Dernière mise à jour il y a %{time} heure",
+      "saved_hours_ago": "Dernière mise à jour il y a %{time} heures",
+      "saved_day_ago": "Dernière mise à jour il y a %{time} jour",
+      "saved_days_ago": "Dernière mise à jour il y a %{time} jours",
```
Crash--: Does it fit well on a very small screen?

edas: We don't display this on small screens yet (we will, but the question will be raised at this time when I will ask for the correct UI for small screens))

src/locales/en.json
```diff
@@ -31,16 +31,16 @@
...
     "SavingIndicator": {
-      "saved_just_now": "Saved just now.",
-      "saved_secs_ago": "Saved %{time} sec ago",
-      "saved_min_ago": "Saved %{time} min ago",
-      "saved_mins_ago": "Saved %{time} min ago",
-      "saved_hour_ago": "Saved %{time} hour ago",
-      "saved_hours_ago": "Saved %{time} hours ago",
-      "saved_day_ago": "Saved %{time} day ago",
-      "saved_days_ago": "Saved %{time} days ago",
+      "saved_just_now": "Last update now",
```
y-lohse: Shouldn't it be `Last updated`? Or `Last update: now` or `last update was now`, but `Last update now` seems wrong.

src/locales/en.json
```diff
@@ -31,16 +31,16 @@
...
+      "saved_secs_ago": "Last update %{time} sec. ago",
+      "saved_min_ago": "Last update %{time} min. ago",
+      "saved_mins_ago": "Last update %{time} min. ago",
+      "saved_hour_ago": "Last update %{time} hour ago",
+      "saved_hours_ago": "Last update %{time} hours ago",
+      "saved_day_ago": "Last update %{time} day ago",
+      "saved_days_ago": "Last update %{time} days ago",
       "saving": "Saving...",
-      "saving_impossible": "Impossible to save.",
+      "saving_impossible": "Saving... ...",
```
y-lohse:
```suggestion
      "saving_impossible": "Saving...",
```

src/locales/fr.json
```diff
@@ -31,16 +31,16 @@
...
       "saving": "Enregistrement en cours...",
-      "saving_impossible": "Enregistrement impossible.",
+      "saving_impossible": "Enregistrement en cours... ...",
```
y-lohse:
```suggestion
      "saving_impossible": "Enregistrement en cours...",
```

src/locales/en.json
edas: not as it is requested : https://trello.com/c/T1FZP8U9/132-etu-jai-linformation-de-lenregistrement-de-ma-note-2j

src/locales/en.json
edas: nope, @joel-costa requires explicit 2 times 3 dots

src/locales/fr.json
edas: nope, @joel-costa requires explicit 2 times 3 dots

src/locales/en.json
y-lohse: Mmh ok, but that is super strange, and IMO not clear at all. At best it makes it look like we're trying to save again which is not what we do AFAIK?

I'm not sure if it's intentional or not, but in the trello card there is no space between the sets of dots, it more 6 dots than 2 sets of 3 dots.

edas:
Vu que la carte a pris du temps, j'ai mis à jour avec l'étape supplémentaire (ajout de l'âge de la dernière sauvegarde quand une sauvegarde est en attente depuis longtemps)

src/lib/utils.spec.js
```diff
@@ -137,3 +138,27 @@ describe('getFolderLink', () => {
     })
   })
 })
+
+describe('relativeAge', () => {
+  // constants
+  const sec = 1000
+  const min = 60 * sec
+  it('should have the recorded response for age=3s', () => {
```
ptbrowne: Nit: "should return the right data for age=3s" is a bit easier to understand (I couldn't understand why "recorded" was used, I looked for a memoize in the implementation)

src/lib/utils.js
```diff
@@ -153,3 +153,64 @@ export function createNoteDocument(client) {
...
+]
+
+/**
+ * @typedef {object} RelativeAge
+ * @property {string} key - translation (sub) key
+ * @property {unit} unit - unit (minute, second, hour…) in which the age will be displayed
+ * @property {integer} coef - in how much values of this unit do we count (like every 10 minutes)
+ * @property {integer} time - age to be inserted in the translation message
+ * @property {integer} interval - `unit * coef`: compute again in that timeframe (in milliseconds)
+ */
```
ptbrowne: :+1:

edas:
New wordings from https://fabien-notes.mycozy.cloud/public/?sharecode=afSztF186uGS#/

PR #116, opened 2020-03-06
feat: Offline Indicator

edas:
When offline, shows a message about it. On large
views, it shows as a tooltip of the saving indicator,
and is forced opened for a few seconds.
On mobile view, it's a banner on top of the page.


src/components/notes/offline-indicator.jsx
```diff
@@ -0,0 +1,90 @@
...
+    <Tooltip
+      title={props.content}
+      open={props.open || undefined}
+      key={props.open ? 'open' : 'auto'}
+      classes={{ tooltip: styles.tooltip }}
+    >
+      {props.children}
+    </Tooltip>
+  )
+}
```
Crash--: Should be moved to UI?

src/components/notes/offline-indicator.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const { t } = useI18n()
+  const { isMobile } = useBreakpoints()
+  const content = (
+    <Grid
+      direction="row"
+      container
+      wrap="nowrap"
+      spacing={16}
+      className={styles.grid}
+    >
```
Crash--: I'm wondering if this Grid should not be moved to UI. You're setting a 1rem padding and just for that, it should be centralized to UI.

src/components/notes/offline-indicator.styl
```diff
@@ -0,0 +1,26 @@
...
+        margin 0
+        width 100%
+
+.chip
+    background-color var(--silver)
+    opacity 60%
+
+.tooltip
+    background-color var(--paleGrey) !important
+    border-radius 8px !important
```
Crash--: Why those important? When I take a look at https://v3.material-ui.com/demos/tooltips/ I can see a `lightTooltip` whiteout any important

src/components/notes/offline-indicator.styl
```diff
@@ -0,0 +1,26 @@
...
+    box-shadow 0 1px 3px 0 rgba(50, 54, 63, 0.19), 0 6px 18px 0 rgba(50, 54, 63, 0.19)
+    color var(--black) !important
+    line-height 1.3 !important
+    position relative
+    left -1rem
```
Crash--: Generally I don't like to have a negative value. Specially when using a tooltip and a popover component. Why do you need it?

src/components/notes/offline-indicator.styl
```diff
@@ -0,0 +1,26 @@
...
+        width 100%
+
+.chip
+    background-color var(--silver)
+    opacity 60%
+
+.tooltip
+    background-color var(--paleGrey) !important
+    border-radius 8px !important
+    box-shadow 0 1px 3px 0 rgba(50, 54, 63, 0.19), 0 6px 18px 0 rgba(50, 54, 63, 0.19)
```
Crash--: should we use variables for colors?

src/components/notes/saving-indicator.jsx
```diff
@@ -2,28 +2,47 @@ import React, { useEffect, useCallback, useState } from 'react'
 import { useI18n } from 'cozy-ui/react/I18n'
 import { Text } from 'cozy-ui/react/Text'
 import usePeriodicRender from 'cozy-ui/react/hooks/usePeriodicRender'
+import useEventListener from 'cozy-ui/react/hooks/useEventListener'
+import useBreakpoints from 'cozy-ui/react/hooks/useBreakpoints'
 import styles from 'components/notes/saving-indicator.styl'
 import throttle from 'lodash/throttle'
 import { relativeAge } from 'lib/utils'
-
+import OfflineIndicator from './offline-indicator'
```
Crash--:
```suggestion
import OfflineIndicator from 'components/notes/offline-indicator'
```

src/components/notes/offline-indicator.styl
```diff
@@ -0,0 +1,26 @@
+.grid
+    padding 1rem
+    font-size 1rem
```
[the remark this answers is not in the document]
edas: The tooltip above sets it to something lower

src/components/notes/offline-indicator.styl
edas: We need to apply some alpha transparency so we can't use our standard variables. We could set a special variable for this but shadows colors don't depends on the theme colors anyways.

src/components/notes/offline-indicator.styl
edas: Their CSS is loaded after ours. I could also use an higher specificity for the selector but I  it doesn't seem really better.

Crash--: Ça me parait difficilement imaginable qu'on ne puisse pas surcharger ce qu'on veuille via leur API CSS ou via les thèmes et qu'on doive passer par des important. On target bien les bons éléments ? Genre on est sur devoir toucher au .tooltip et non au  .popper ?

src/components/notes/offline-indicator.styl
ptbrowne: I think it is a good idea to use a variable for shadows : https://github.com/cozy/cozy-ui/blob/master/stylus/settings/shadows.styl

ptbrowne:
Ping :)

src/components/notes/offline-indicator.jsx
edas: I am not convinced. This is the first use, and I really have no clue if other uses of a grid may need a 1rem padding. I would wait for a second or third use before trying to mutualize a technical component

src/components/notes/offline-indicator.styl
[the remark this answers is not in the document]
edas: migrated to tooltip

src/components/notes/offline-indicator.styl
edas: je n'ai pas réussi à changer l'ordre d'insertion des CSS de MUI. Je suis passé par le CSS-in-JS du coup

src/components/notes/offline-indicator.styl
edas: Je suis passé autrement (vu que j'ai utilisé le CSS in JS plus haut, j'ai pu changer le positionnement à la source)

src/components/notes/offline-indicator.styl
edas: J'ai du passer en CSS-in-JS pour ça, et du coup je casse la possibilité d'utiliser stylus

src/components/notes/offline-indicator.jsx
edas: Je suis dessus

edas:
Should be ok now. I will update the cozy-ui version once the Tooltip is merged

src/components/notes/saving-indicator.spec.js
```diff
@@ -33,12 +49,27 @@ function itMatchSnapshot(collabProvider) {
...
 describe('SavingIndicator', () => {
-  afterEach(() => jest.resetAllMocks())
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+  beforeEach(() => {
+    useBreakpoints.mockImplementation(() => ({
+      isMobile: false
+    }))
+    global.document.createRange.mockImplementation(() => ({
```
ptbrowne: Nit: A comment stating why document.createRange could be useful (at first I thought it was the same as lodash's range function, I forgot about this name for selections).

src/components/notes/saving-indicator.jsx
```diff
@@ -2,28 +2,47 @@ import React, { useEffect, useCallback, useState } from 'react'
...
+    window && window.navigator && window.navigator.onLine === false
+  )
+  const setNavigatorOffline = useCallback(() => setIsNavigatorOffline(true), [
+    setIsNavigatorOffline
+  ])
+  const setNavigatorOnline = useCallback(() => setIsNavigatorOffline(false), [
+    setIsNavigatorOffline
+  ])
+  useEventListener(window, 'online', setNavigatorOnline)
+  useEventListener(window, 'offline', setNavigatorOffline)
```
ptbrowne: Nit: I am wondering if some of the online/offline logic could be moved outside the component ? Maybe with a useNetworkStatus hook ? It could also take the offlineSince state part ?

edas: I don't think we'll really mutualize anything and it will add a custom extra layer upon something simple and already known

PR #119, opened 2020-03-17
feat: Initial of ? for unknown public user

edas:
For a share by link, if the visitor does not provide an
explicit username, use '?' as initial for his cursors.

### DIFF
--- a/src/hooks/useUser.js
+++ b/src/hooks/useUser.js
@@ -1,13 +1,14 @@
-import { useMemo } from 'react'
-
+import { useMemo, useContext } from 'react'
+import IsPublicContext from 'components/IsPublicContext'
 import { getShortNameFromClient, getUserNameFromUrl } from 'lib/utils.js'
 
 function useUser({ userName: providedUserName, cozyClient }) {
+  const isPublic = useContext(IsPublicContext)
   const userName = useMemo(
     () =>
       providedUserName ||
       getUserNameFromUrl() ||
-      getShortNameFromClient(cozyClient),
+      (isPublic ? '?' : getShortNameFromClient(cozyClient)),
     [cozyClient, providedUserName]
   )
   const userId = userName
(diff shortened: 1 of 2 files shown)


src/hooks/useUser.spec.js
```diff
@@ -1,17 +1,39 @@
...
       expect(result.current.userName).toEqual('hello world')
     })
   })
+
+  describe('for a public view', () => {
+    it('shows an unknown initial', () => {
+      const context = withPublicContext(true)
+      setUrlUsername(null)
+      const { result } = renderHook(() => useUser({}), { wrapper: context })
+      expect(result.current.userName[0]).toEqual('?')
```
Crash--: why testing only the first char of the string? We should only have one char,no ?

src/hooks/useUser.js
```diff
@@ -1,13 +1,14 @@
...
 import { getShortNameFromClient, getUserNameFromUrl } from 'lib/utils.js'
 
 function useUser({ userName: providedUserName, cozyClient }) {
+  const isPublic = useContext(IsPublicContext)
   const userName = useMemo(
     () =>
       providedUserName ||
       getUserNameFromUrl() ||
-      getShortNameFromClient(cozyClient),
+      (isPublic ? '?' : getShortNameFromClient(cozyClient)),
```
Crash--: Hmmm... On ne peut pas être dans un contexte public tout en ayant un username en params dans l'URL ?

src/hooks/useUser.spec.js
edas: Probably, but as the spec from Zoe defines only the initial for the cursor, I wanted the test to reflect the spec and only the spec

src/hooks/useUser.js
edas: Si, en venant de Desktop ou de Drive, ça peut arriver (d'où la ligne 10)

PR #121, opened 2020-03-17
Mobile UI enhancements

y-lohse:
The UI of notes on mobile is close to broken, this PR gets things to an acceptable level 😅

Before:

Capture d'écran 2020-03-17 14 35 39
Capture d'écran 2020-03-17 16 01 11

After:

Capture d'écran 2020-03-17 15 56 03
Capture d'écran 2020-03-17 16 01 45


Crash--:
How do we display the fact that a note is shared?

y-lohse:
On mobile not at all at the moment. Not sure what the final design should look like.

PR #122, opened 2020-03-19
fix: Edit a note on the correct cozy instance

edas:
Previously, clicking on a note in the index always redirect
to the editor's instance. This was creating duplicates and
conflicts when the note was created on another instance.
It now correctly redirects to the first author's instance.

### DIFF
--- a/src/lib/utils.spec.js
+++ b/src/lib/utils.spec.js
@@ -3,8 +3,19 @@ import {
   fetchIfIsNoteReadOnly,
   getFolderLink,
   relativeAge,
-  getUserNameFromUrl
+  getUserNameFromUrl,
+  generateReturnUrlToNotesIndex
 } from './utils'
+
+// see https://remarkablemark.org/blog/2018/11/17/mock-window-location/
+const { location } = window
+function setLocation(url) {
+  delete window.location
+  window.location = url
+}
+function restoreLocation() {
+  window.location = location
+}
 
 function setupClient(verbs = [], ids = ['first', 'other']) {
   return {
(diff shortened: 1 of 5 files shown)


src/lib/utils.spec.js
```diff
@@ -178,3 +189,53 @@ describe('getUserNameFromUrl', () => {
...
+    }
+  }
+
+  function createNote(id = '12345') {
+    return { id, type: 'io.cozy.files' }
+  }
+
+  afterEach(() => {
+    restoreLocation()
+  })
```
y-lohse: Nit: I find it strange that we restore location after each test, but only alter it in some of them. Either we want the same behavior all the time and it would make sense to call `setLocation` in a `beforeEach`, or it's on-demand and we can call `restoreLocation` in the tests that need it?

src/lib/utils.spec.js
```diff
@@ -178,3 +189,53 @@ describe('getUserNameFromUrl', () => {
...
+
+  it('returns an URL string', async () => {
+    const client = createClient()
+    const note = createNote()
+    const urlString = await generateReturnUrlToNotesIndex(client, note)
+    const url = new URL(urlString)
+    expect(url.toString()).toEqual(urlString)
```
y-lohse: When can this test fail? SInce we compare the result of the function with itself, I don't think it can? It probably covers the case where `new URL()` fails wut it's not very explicit.

src/lib/utils.spec.js
edas: Restoring location is harmless, even if you don't setLocation before. A afterEach allows to not forget anything.

src/lib/utils.spec.js
edas: Yep, it'll fail if `url` is not an URL. That was the purpose

PR #124, opened 2020-03-23
fix: Saving indicator

edas:
We add errors in the saving indicator messages, this changes fixes these.
I also tried to refactor the code, which became too complicated.


src/components/notes/saving-indicator.jsx
```diff
@@ -19,6 +20,61 @@ const outOfSyncAfter = 30 * sec
...
+}
+
+/**
+ * Checks for disconnection (browser offline or out of sync)
+ *
+ * @param {object} params
+ * @param {boolean} params.isOutOfSync
+ * @returns {{isDisconnected, disconnectedAge}}
+ */
+function checksDisconnection({ isOutOfSync }) {
```
ptbrowne: Since it is a hook, I think it would be better to keep the `use` convention. useDisconnection ?

src/components/notes/saving-indicator.jsx
```diff
@@ -19,6 +20,61 @@ const outOfSyncAfter = 30 * sec
...
+  const isDisconnected = !!disconnectedSince
+  useEffect(() => {
+    const disconnectedNow = isBrowserOffline || isOutOfSync
+    if (!!disconnectedNow !== !!isDisconnected) {
+      setDisconnectedSince(disconnectedNow ? new Date() : null)
+    }
+  }, [isBrowserOffline, isOutOfSync, setDisconnectedSince, isDisconnected])
+  const disconnectedAge = isDisconnected ? now - disconnectedSince : 0
+  return { isDisconnected, disconnectedAge }
+}
```
ptbrowne: Nit: I don't understand why we have isOutOfSync in this function ? Do you think it could be "mixed" in later by the caller ? I think having both concepts "offline" and "out of sync" is a bit confusing.

src/components/notes/saving-indicator.jsx
```diff
@@ -19,6 +20,61 @@ const outOfSyncAfter = 30 * sec
...
+ * Should we open the offline indicator tooltip?
+ *
+ * We open it for a few seconds when we go offline
+ * then close it (it is still accessible when hovering over)
+ * @param {object} params
+ * @param {integer} params.disconnectedAge - milliseconds since when we are disconnected
+ * @param {boolean} params.isDisconnected
+ * @returns {boolean}
+ */
+function shouldOpenOfflineIndicator({ disconnectedAge, isDisconnected }) {
```
ptbrowne: I think we should use the hook naming convention here.

src/components/notes/saving-indicator.jsx
```diff
@@ -19,6 +20,61 @@ const outOfSyncAfter = 30 * sec
...
+  }, [isBrowserOffline, isOutOfSync, setDisconnectedSince, isDisconnected])
+  const disconnectedAge = isDisconnected ? now - disconnectedSince : 0
+  return { isDisconnected, disconnectedAge }
+}
+
+/**
+ * Should we open the offline indicator tooltip?
+ *
+ * We open it for a few seconds when we go offline
+ * then close it (it is still accessible when hovering over)
```
ptbrowne: I find the usePeriodicRender used here, a bit cryptic. I think using an imperative API like Alert here may have been simpler.

src/components/notes/saving-indicator.jsx
```diff
@@ -27,80 +83,60 @@ const outOfSyncAfter = 30 * sec
...
     } else {
       message = t('Notes.SavingIndicator.saving')
     }
   } else {
-    message = t(`Notes.SavingIndicator.saved.${key}`, { time })
+    if (isDisconnected) {
+      message = t(`Notes.SavingIndicator.out_of_sync.${key}`, { time })
+    } else {
+      message = t(`Notes.SavingIndicator.saved.default`, { time })
+    }
```
ptbrowne: Nit: The message building could be extracted. This would simplify the render.

src/components/notes/saving-indicator.jsx
edas: I use the following vocabulary : 

- Offline: the browser is offline, or declares to be offline
- Out of sync: for an unknown reason (no connection, error on the server, or anything else), we couldn't commit our change for a long time
- Disconnected: We know we are not correctly connected to the server (either offline or out of sync)

This hook/function is for the third case, and we need it to remember the date in which we enter this state

src/components/notes/saving-indicator.jsx
edas: We could, but I found the name not really clear, and as the function is in the same file, I did think it wasn't as important.

src/components/notes/saving-indicator.jsx
edas: What would you expect ?

ptbrowne: For this kind of things, I am accustomed to doing things like:

```
if (propertyHasChanged) {
  showAlert('Something has changed', 5000)
}
```

- Here I find it a bit convoluted to use a usePeriodicRender to remember to re-render at the end of the display period for the alert to disappear. I find that it's low-level to care about `render` here.

- I find the imperative API more suited for fire and forget alerts

src/components/notes/saving-indicator.jsx
ptbrowne: Nit: I wonder if this "remember-the-date-in-which-we-enter-this-state" could not be expressed with a dedicated hook ?

```
const useAge = value => {
  const [lastValue, setLastValue] = useState(value)
  const [lastTime, setLastTime] = useState(Date.now())
  useEffect(() => {
    if (value !== lastValue) {
      setLastTime(Date.now())
      setLastValue(value)
    }
  }, [value, lastValue, setLastValue])
  return lastTime
}

const useDisconnection = ({ isOutOfSync }) => {
  const isBrowserOffline = useBrowserOffline()
  const isDisconnected = isBrowserOffline || isOutOfSync
  const disconnectedValueAge = useAge(isDisconnected)
  const disconnectedAge = isDisconnected ? disconnectedValueAge : 0
  return { isDisconnected, disconnectedAge }
}
```

edas: I was doing that this morning, and found something way simplier which probably doesn't need any helper : 

```
const disconnectedSince = useMemo(() => new Date(), [isDisconnected])
```
which gives
```
function useDisconnection({ isOutOfSync }) {
  const now = new Date()
  const isBrowserOffline = useBrowserOffline()
  const isDisconnected = isBrowserOffline || isOutOfSync
  const disconnectedSince = useMemo(() => new Date(), [isDisconnected])
  const disconnectedAge = isDisconnected ? now - disconnectedSince : 0
  return { isDisconnected, disconnectedAge }
}
```

PR #125, opened 2020-03-23
fix: Notes creation on mobile

edas:
The update of cozy-client broke the "+" button on mobile.
The AddMobile component uses the withClient / useClient of the
new cozy-client, which relies on the new Context feature of React.
The bar only provides the old Context feature in its tree.
When the AddMobile is inserted into the bar, it cannot find the
cozyClient instance in the tree, and fails.

### DIFF
--- a/src/components/notes/add.jsx
+++ b/src/components/notes/add.jsx
@@ -1,14 +1,16 @@
 import React, { useState, useCallback } from 'react'
 
-import { withClient } from 'cozy-client'
+import { useClient } from 'cozy-client'
 
 import Button from 'cozy-ui/react/Button'
-import { translate } from 'cozy-ui/react/I18n'
+import { useI18n } from 'cozy-ui/react/I18n'
 import BarButton from 'cozy-ui/react/BarButton'
 
 import { createNoteDocument, generateReturnUrlToNotesIndex } from 'lib/utils'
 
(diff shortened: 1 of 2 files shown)


src/components/notes/add.jsx
```diff
@@ -1,14 +1,16 @@
...
+import { useClient } from 'cozy-client'
 
 import Button from 'cozy-ui/react/Button'
-import { translate } from 'cozy-ui/react/I18n'
+import { useI18n } from 'cozy-ui/react/I18n'
 import BarButton from 'cozy-ui/react/BarButton'
 
 import { createNoteDocument, generateReturnUrlToNotesIndex } from 'lib/utils'
 
-const Add = ({ t, className, client }) => {
```
ptbrowne: I see that t and client were removed from props, but I don't see the call site to see if those props were removed ? Was it wrapped in withClient or translate before ?

edas: yep, `t` and `client` where from withClient() and translate() 

See https://github.com/cozy/cozy-notes/pull/125/files#diff-e70fe36241ca5b019a08946e7d13bfe3L42

ptbrowne: Ah ok, sorry, since the export default had changed of place, I did not connect the dots.

PR #145, opened 2020-06-02
Editable/uneditable shared notes

y-lohse:
This PR updates cozy-sharing to the latest version where instead of a hard-coded behavior for Notes, the user can choose if a note should be shared with or without editing permissions.

We also added a small indicator about a note being uneditable when that's the case:

Capture d'écran 2020-06-02 10 59 34

Requires https://github.com/cozy/cozy-libs/pull/1015

### DIFF
--- a/src/components/notes/EditorCorner.jsx
+++ b/src/components/notes/EditorCorner.jsx
@@ -1,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Icon from 'cozy-ui/react/Icon'
+import Tooltip from 'cozy-ui/react/Tooltip'
+
+import SharingWidget from 'components/notes/sharing'
+
+const EditorCorner = ({ doc, isPublic, isReadOnly }) => {
+  if (!isPublic) return <SharingWidget file={doc.file} />
+  else if (isReadOnly)
+    return (
+      <Tooltip title={'This note is in read-only mode.'}>
+        <Icon icon="lock" color="var(--primaryTextColor)" />
+      </Tooltip>
+    )
+  else return false
+}
+
+EditorCorner.propTypes = {
+  doc: PropTypes.object.isRequired,
+  isPublic: PropTypes.bool.isRequired,
+  isReadOnly: PropTypes.bool.isRequired
+}
+
+export default EditorCorner
--- a/src/components/notes/EditorCorner.spec.jsx
+++ b/src/components/notes/EditorCorner.spec.jsx
@@ -1,0 +1,45 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import SharingWidget from 'components/notes/sharing'
+import EditorCorner from './EditorCorner'
+
+describe('EditorCorner', () => {
+  const mockDoc = {}
+  it('shows the sharing widget on private views', () => {
+    const readOnlyComponent = shallow(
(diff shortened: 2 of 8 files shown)


src/components/notes/EditorCorner.spec.jsx
```diff
@@ -0,0 +1,45 @@
...
+    )
+    expect(readOnlyComponent.find(SharingWidget).length).toBe(1)
+
+    const editableComponent = shallow(
+      <EditorCorner doc={mockDoc} isPublic={false} isReadOnly={false} />
+    )
+    expect(editableComponent.find(SharingWidget).length).toBe(1)
+  })
+
+  it('shows the editor is read only on public views', () => {
```
Crash--, line 21:
```suggestion
  it('shows the editor corner on read only public views', () => {
```

y-lohse: Mmh en fait je voulais dire "indique que l'éditeur est en read-only sur la vue publique", c'est pas suffisamment clair?

Crash--: Bah c'est pas ce que ça fait. Tu shallow EditorCorner pas Editor. Du coup tu testes que l'éditeur corner dispose d'une icone lock et d'un titre "read only" quand c'est isPublic et isReadOnly.

src/components/notes/EditorCorner.jsx
```diff
@@ -0,0 +1,26 @@
...
+import Tooltip from 'cozy-ui/react/Tooltip'
+
+import SharingWidget from 'components/notes/sharing'
+
+const EditorCorner = ({ doc, isPublic, isReadOnly }) => {
+  if (!isPublic) return <SharingWidget file={doc.file} />
+  else if (isReadOnly)
+    return (
+      <Tooltip title={'This note is in read-only mode.'}>
+        <Icon icon="lock" color="var(--coolGrey)" />
```
ptbrowne, line 14: Nit: It would good not to have to hard code the icon color here. This hardcoded color would not work for example if we code a dark mode. Maybe we could achieve this with a prop `textIcon` that would set its color to `var(--iconTextColor)` that would be set by default to `var(--coolGrey)` ?

See https://github.com/cozy/cozy-ui/issues/1475 also.

src/components/notes/sharing.jsx
```diff
@@ -1,35 +1,33 @@
...
 import { withClient } from 'cozy-client'
 
 import useFileWithPath from 'hooks/useFileWithPath'
 import styles from 'components/notes/sharing.styl'
 
-export default withLocales(
-  withClient(function SharingWidget(props) {
-    const { client } = props
+export default withClient(function SharingWidget(props) {
+  const { client } = props
```
ptbrowne, line 9: Nit: useClient could be used here.

src/components/notes/EditorCorner.spec.jsx
y-lohse: C'est vrai, je vais éditer le message.

src/components/notes/EditorCorner.jsx
y-lohse: Good point, I ended up using `primaryTextColor` because the icon has no text next to it by default, but `iconTextColor ` is a good idea re. the cozy-ui issue.

y-lohse:
Damnit, I merged it before updating cozy-sharings again 🤦‍♂️

PR #147, opened 2020-06-04
fix: Use tranlated string for read-only mode

y-lohse:
This was introduced in #145, the string was hard-coded (but it was the right text instead of a placeholder, so I missed it 😬)

### DIFF
--- a/src/components/notes/EditorCorner.jsx
+++ b/src/components/notes/EditorCorner.jsx
@@ -3,18 +3,20 @@ import PropTypes from 'prop-types'
 
 import Icon from 'cozy-ui/react/Icon'
 import Tooltip from 'cozy-ui/react/Tooltip'
+import { useI18n } from 'cozy-ui/react/I18n'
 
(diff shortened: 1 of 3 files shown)


src/components/notes/EditorCorner.jsx
```diff
@@ -3,18 +3,20 @@ import PropTypes from 'prop-types'
...
+  else if (isReadOnly) {
+    const { t } = useI18n()
     return (
-      <Tooltip title={'This note is in read-only mode.'}>
+      <Tooltip title={t('Notes.Editor.read_only')}>
         <Icon icon="lock" color="var(--primaryTextColor)" />
       </Tooltip>
     )
-  else return false
+  } else return false
```
Crash--, line 19: Don"t we return null usually?

y-lohse: Man I can never remember if it's null or false. But you're right it's null.

PR #158, opened 2020-07-23
chore: Bump version

y-lohse:


ptbrowne:
LGTM. (I think there is no need for reviews for those kind of commits :-) )

y-lohse:
Me neither, that's why I didn't request any 😁 but thanks anyway haha

ptbrowne:
Héhé fair point.

PR #166, opened 2020-11-05
Feat/upgrade

Crash--:
commits should be self explanatory but the main objective was to upgrade cozy-client / ui & sharing


package.json
```diff
@@ -55,14 +55,14 @@
     "@material-ui/core": "3.9.4",
     "@material-ui/styles": "4.10.0",
     "cozy-bar": "7.7.8",
-    "cozy-client": "16.0.0",
+    "cozy-client": "16.1.1",
```
ptbrowne, line 57: With gitup "F", it is easy to fixup commits ;)

src/targets/browser/index.jsx
```diff
@@ -128,3 +128,8 @@ export const initApp = () => {
 }
 
 document.addEventListener('DOMContentLoaded', initApp)
+
+if (module.hot) {
+  initApp()
```
ptbrowne, line 133: A new client is created, are we sure we want that ?

ptbrowne (approved):

package.json
Crash--: oublié de squasher... j'ai bien le fixup en plus... Ça m'apprendra à pas relire les commits en soumettant

src/components/notes/EditorCorner.spec.jsx
```diff
@@ -10,13 +13,28 @@ import EditorCorner from './EditorCorner'
 describe('EditorCorner', () => {
   const mockDoc = {}
   it('shows the sharing widget on private views', () => {
-    const readOnlyComponent = shallow(
-      <EditorCorner doc={mockDoc} isPublic={false} isReadOnly={true} />
+    //  useI18n.mockReturnValue({})
```
ptbrowne, line 16: Can you remove the comment ?

src/components/notes/EditorCorner.spec.jsx
```diff
@@ -10,13 +13,28 @@ import EditorCorner from './EditorCorner'
...
+    const readOnlyComponent = mount(
+      <EditorCorner doc={mockDoc} isPublic={false} isReadOnly={true} />,
+      {
+        wrappingComponent: I18n,
+        wrappingComponentProps: {
+          lang: 'en',
+          dictRequire: () => en
+        }
+      }
```
ptbrowne, line 25: Nit: It is useful to make an AppLike component containing all the wrappers necessary for tests.

ptbrowne (approved):
👍 🎉

src/components/notes/EditorCorner.spec.jsx
Crash--: Thanks!

PR #167, opened 2020-12-04
Feat/upgrade atlaskit

Crash--:
Commits & comments should be enough to understand what is going on on this PR!

### DIFF
--- a/src/lib/collab/schema.js
+++ b/src/lib/collab/schema.js
@@ -1,26 +1,49 @@
 import { Schema } from 'prosemirror-model'
 
-// taken from a debug of @atlakit/editor/editor-core/create-editor/create-editor
-// L139 (new Schema({nodes ,marks}))
 // static because the @atlaskit code base requires a real navigator
 // TODO: either find and exclude plugins requiring interaction
 //       or running a JSDOM faking a navigator
+// @quentin: I don't understand yet why we need to do that, so if someone
+// knows, then update the comment ;)
 
-const nodes = [
+// To get the schema, we need to log the values from node_modules/@atlaskit/editor/editor-core/create-editor/create-schema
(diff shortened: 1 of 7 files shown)


Crash--:
@nono I asked you to review this PR so that you would be aware of the change in the Schema.

src/lib/collab/schema.js
```diff
@@ -1,26 +1,49 @@
...
+
+//https://github.com/cozy/cozy-notes/pull/2/commits/202e529cdb1e71996c2cab43057984c9f885b61a
+
+export const nodes = [
   [
     'doc',
     {
       content: '(block)+',
-      marks: 'link'
+      marks: 'link unsupportedMark unsupportedNodeAttribute'
```
nono, line 36: Do you know what are `unsupportedMark` and `unsupportedNodeAttribute`?

Crash--: Nop. The changelog is here https://atlaskit.atlassian.com/packages/editor/editor-core/changelog/%3E=114 I didn't take the time to read it yet

Crash--: It seems that is there to preserve nodes & marks even if they are not supported by the fragment

src/lib/collab/schema.js
```diff
@@ -1,26 +1,49 @@
...
     }
   ],
   [
     'paragraph',
     {
+      selectable: false,
```
ptbrowne, line 42: Why do we need to put selectable: false ?

ptbrowne (approved):

src/lib/collab/schema.js
Crash--: I don't know... yet. To be honest, I don't know why we need to dump this internal schema from atlaskit yet.

PR #175, opened 2021-01-12
Feat/upgrade schema

Crash--:
Add to Notes the capability to upgrade a schema for a Note. Useful when we add new plugins or upgrade Altaskit plugins.

- Store the schema's version in `metadata.schema.version` when we create a note 
- If the schema of the note is different from the one of the app, we call the stack to update the schema 
- Handle the case of a shared note using realtime to reload the app

### DIFF
--- a/src/lib/collab/provider.js
+++ b/src/lib/collab/provider.js
@@ -70,6 +70,7 @@ export class CollabProvider {
     })
     this.channel.on('data', this.onReceiveData)
     this.channel.on('telepointer', this.onReceiveTelepointer)
+    this.channel.on('schemaupdated', this.onReceiveSchemaUpdated)
     this.channel.on('needcatchup', () => this.catchup())
     const state = getState()
     const doc = jsonTransformer.encode(state.doc)
@@ -285,7 +286,13 @@ export class CollabProvider {
     this.updateParticipant(sessionId, data.timestamp)
     this.emit('telepointer', data)
   }
-
+  /**
+   * We receive an update of the schema from the server
+   * We reload the page
+   */
+  onReceiveSchemaUpdated = () => {
+    window.location.reload(true)
+  }
   updateParticipant(sessionId, timestamp) {
     const userId = this.serviceClient.getUserId(sessionId)
     const participant = getParticipant({ userId, sessionId })
--- a/src/lib/collab/schema.js
+++ b/src/lib/collab/schema.js
@@ -28,6 +28,14 @@ import { Schema } from 'prosemirror-model'
 
 //https://github.com/cozy/cozy-notes/pull/2/commits/202e529cdb1e71996c2cab43057984c9f885b61a
 
+// if you edit the schema, please upgrade this schemaVersion
+
+export const schemaVersion = 1
+
+export const getSchemaVersion = () => {
+  return schemaVersion
+}
+
 export const nodes = [
   [
     'date',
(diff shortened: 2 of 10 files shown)


src/lib/collab/schema.js
```diff
@@ -743,6 +743,9 @@ export const schemaObject = {
 
 export const schemaVersion = 1
 
+export const getSchemaVersion = () => {
+  return schemaVersion
+}
```
JF-Cozy, line 748: Pourquoi faire une fonction qui retourne une const et pas directement importer la const (à la place de faire `getSchemaVersion()`) ?

Crash--: Parce que je sais mocker une fonction mais pas une const sans devoir retoucher à la structure de l'export de ce fichier. 

Mais merci de le souligner, il faut que je remonte cette var `schemaVersion` en haut de fichier à côté du schéma et dire qu'il faut l'upgrader à chaque fois qu'on y touche

JF-Cozy: Ah oui ok je me doutais qu'il y avait une astuce mais je voyais pas laquelle...

src/lib/collab/provider.js
```diff
@@ -285,7 +286,13 @@ export class CollabProvider {
     this.updateParticipant(sessionId, data.timestamp)
     this.emit('telepointer', data)
   }
-
+  /**
+   * We receive an update of the schema from the server
+   * We reload the page
+   */
+  onReceiveSchemaUpdated = () => {
+    window.location.reload(true)
```
ptbrowne, line 294: Should we check that all the modifications have been saved ? Can the user lose data if the schema update happen while she is modifying the document ?

src/lib/collab/stack-client.js
```diff
@@ -146,7 +175,10 @@ export class ServiceClient {
    */
   onRealtimeEvent(doc) {
     const id = doc.id || doc._id
-    const type = doc.doctype
+    let type = doc.doctype || doc.type
+    // Hack needed since realtime data from io.cozy.notes & io.cozy.files
+    // are not structured the same way
+    if (type === 'file') type = 'io.cozy.files'
```
ptbrowne, line 181: In the future, should this be normalized via cozy-realtime ?

ptbrowne (approved):
LGTM

src/lib/collab/provider.js
Crash--: I tried to explain that in the commit message 

"There is a potential risk of losing the latest editions from Alice (if
Alice network is slow the realtime event can be slow to arrive) but we
accept it since this case should not happen very often."

ptbrowne: Oups sorry, I checked all the changes at once this time 😅  Thanks for the explanation.

Crash--: I learned from someone that writing a good commit message is essential, I follow this advice! I can now have more lines in a commit message than in the code itself. 

To be more precise about the shortcut we took, we have a discussion about a potential solution to reduce at the minimum the risk of loosing data but it requires more work (and need to be more challenged), something like: 
- When we detect from the opening event that the schema change, before updating the schema, send a message to the channel saying "make a save, I'll refresh you in a second". Then all subscribers should send their latest steps, waits for reconciliation if needed, and then send back "i'm ready". When we receive all these messages, then we can apply the migration and the refresh. 

We decided that for now, the way I handle it in this PR, is enough.

PR #181, opened 2021-03-02
Feat/upgrade cozy libs

Crash--:
- Upgrade to MUI v4
- Upgrade to CC 18


JF-Cozy:
@Crash-- ok pour moi mais CI red

PR #183, opened 2021-03-22
fix: Upgrade sharing to fix issue with ReadOnly

Crash--:
Upgrade Cozy-sharing to get this fix https://github.com/cozy/cozy-libs/commit/f1ebea9

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "cozy-flags": "2.1.0",
     "cozy-realtime": "3.7.0",
     "cozy-scripts": "5.1.1",
-    "cozy-sharing": "2.10.0",
+    "cozy-sharing": "3.3.1",
     "cozy-ui": "45.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "cozy-flags": "2.1.0",
     "cozy-realtime": "3.7.0",
     "cozy-scripts": "5.1.1",
-    "cozy-sharing": "2.10.0",
+    "cozy-sharing": "^3.3.1",
```
ptbrowne, line 63: Is it expected that the dependency is not pinned ? Just to be sure, the breaking change that triggered 2 to 3 is not problematic in notes ?

ptbrowne (dismissed):

package.json
Crash--: Nop, already migrated to mui-v4 so not affected by the BC and thanks for the pin!

PR #201, opened 2021-08-06
fix: update app init

acezard:
Post breaking changes from dependencies

Cozy-flags introduced breaking changes which were not detected earlier. This has been fixed in this mr. Also the dependency tree has been cleaned a bit.

All dependencies in the project are now up to date to their latest minor versions (except two).

No dependency has been updated to its next major version though as it seems risky to do everything at once.

### DIFF
--- a/src/lib/debug.jsx
+++ b/src/lib/debug.jsx
@@ -1,8 +1,9 @@
 import { useEffect } from 'react'
-import { useFlag } from 'cozy-flags'
+import useFlag from 'cozy-flags/dist/useFlag'
+import FlagSwitcher from 'cozy-flags/dist/FlagSwitcher'
 import set from 'lodash/set'
 import get from 'lodash/get'
-import flag, { FlagSwitcher } from 'cozy-flags'
+import flag from 'cozy-flags'
 
 import { getVersion, sendableSteps } from 'prosemirror-collab'
 
--- a/src/targets/browser/index.ejs
+++ b/src/targets/browser/index.ejs
@@ -1,32 +1,45 @@
 <!DOCTYPE html>
 <html lang="{{.Locale}}">
-<head>
-  <meta charset="utf-8">
-  <title><%= htmlWebpackPlugin.options.title %></title>
-  <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
-  <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32">
-  <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16">
-  <link rel="manifest" href="/manifest.json" crossorigin="use-credentials">
-  <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#297EF2">
-  <% if (__TARGET__ !== 'mobile') { %>
-  <link rel="stylesheet" type="text/css" href="//{{.Domain}}/assets/fonts/fonts.css">
-  <% } %>
-  <meta name="theme-color" content="#ffffff">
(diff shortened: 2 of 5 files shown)


src/lib/debug.jsx
```diff
@@ -1,8 +1,9 @@
 import { useEffect } from 'react'
-import { useFlag } from 'cozy-flags'
+import useFlag from 'cozy-flags/dist/useFlag'
```
acezard, line 2: Not a good look but could not get it imported any other way

src/targets/browser/index.ejs
```diff
@@ -1,32 +1,45 @@
...
+      content="width=device-width, height=device-height, initial-scale=1, viewport-fit=cover"
+    />
+    <% _.forEach(htmlWebpackPlugin.files.css, function(file) { %>
+    <link rel="stylesheet" href="<%- file %>" />
+    <% }); %> <% _.forEach(htmlWebpackPlugin.files.js, function(file) { %>
+    <script src="<%- file %>" defer></script>
+    <% }); %> <% if (__TARGET__ === 'mobile') { %>
+    <meta name="format-detection" content="telephone=no" />
+    <script src="cordova.js" defer></script>
+    <% } else { %>
```
acezard, line 40: this is the only non-prettier change here

src/lib/debug.jsx
```diff
@@ -1,8 +1,9 @@
...
+import FlagSwitcher from 'cozy-flags/dist/FlagSwitcher'
```
acezard, line 3: Same as above

src/targets/browser/index.ejs
Crash--: Why did you remove the __STACK_ASSETS__ test? 

Since cozy-bar is still in 7.X, I don't think there is a BC somewhere

acezard: Somehow it prevented the bar from loading, not sure why?

Crash--: This is not the right fix then. 

What is the error when you revert it?

acezard: window.cozy.bar === undefined

acezard: maybe it was a side-effect of cozy-flags, let me try again

acezard: image

nope still broken even after having fixed cozy-flags

acezard: linked to this? 
image

src/lib/debug.jsx
JF-Cozy: Est-ce qu'on pourrait pas regarder du côté de cozy-flag si on peut corriger le souci ?

acezard: Possible oui mais cette MR est assez urgente car master est broken. @Crash-- que préfères-tu ?

Crash--: Créons l'issue sur le répo de cozy-flag et on traitera ça plus tard

Crash--: C'est vraiment important cette notion de si on a un problème suite à une maj de lib de faire l'issue sur la lib pour : 
- tracker le souci 
- pouvoir le corriger
- partager le fait qu'il y a un souci 

Garder ça dans son coin c'est pas productif pour cozy, car on tombera toujours sur le même souci et on reperdra encore et encore du temps sur le sujet

Crash-- (approved):

src/lib/debug.jsx
acezard: issue created https://github.com/cozy/cozy-libs/issues/1330

PR #202, opened 2021-08-09
feat: use updated linter and strict typing

acezard:
This merge request allows us to use the latest eslint features, which is especially important for Typescript. 

In order to make it work, it was necessary to use a fork of cozy-scripts, but also to update the dev dependencies of cozy-notes itself. This is intended as a temporary but necessary change. Indeed, it is difficult to work without proper linting of Typescript files.

It should not in any way introduce regressions in the app itself, because it doesn't touch anything runtime-related. If anything, this configuration will allow us to spot issues before they are delivered to the master branch.


package.json
```diff
@@ -64,7 +67,7 @@
     "cozy-editor-core": "134.0.3",
     "cozy-flags": "2.7.2",
     "cozy-realtime": "3.13.0",
-    "cozy-scripts": "5.10.0",
+    "cozy-scripts": "npm:anc-scripts@^6.0.0",
```
Crash--, line 70: Nein. We don't use personal fork. If you want to use a new version of cozy-scripts or else, you can publish a new tag for our cozy-scripts lib and use it here. You can create a `cozy-script@eslint7" or something

Crash-- (changes requested):
see comment

package.json
acezard: Yes I was expecting that, will do

acezard: updated

Crash-- (approved):

PR #203, opened 2021-08-09
fix: force editor-tables resolution to editor-core

acezard:
https://trello.com/c/HUipo7Qa/50-retour-4-regression-je-narrive-plus-%C3%A0-cr%C3%A9er-des-tableaux-dans-les-notes-erreur-en-console-uncaught-typeerror-e-is-undefined

editor-core as we use it (v134.0.1) will break with editor-tables > 1.1.3. It was necessary to update the resolution of this dependency. I don't think this is a long term solution, best would be to update editor-core itself but unsure at the moment what problems would it create with cozy-stack

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "utf-8-validate": "5.0.5"
   },
   "resolutions": {
-    "prosemirror-model": "1.11.0"
+    "prosemirror-model": "1.11.0",
+    "cozy-editor-core/@atlaskit/editor-tables": "1.1.3"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -459,10 +459,10 @@
     "@babel/runtime" "^7.0.0"
     styled-components "^3.2.6"
 
-"@atlaskit/editor-tables@^1.1.0":
-  version "1.1.5"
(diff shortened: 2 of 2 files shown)


Crash--:
I think this is a good way to handle our editor ;).

Also, when reading the lock file friday, it seems that we have a lot of duplicate packages. You can use DuplicatePackageCheckerPlugin and create a manuel resolution. They do that in banks https://github.com/cozy/cozy-banks/blob/f48951d8c4f89240da782611d1265b9f5a67bb24/config/webpack.config.manual-resolves.js#L28

Crash-- (approved):

acezard:
> I think this is a good way to handle our editor ;).
> 
> Also, when reading the lock file friday, it seems that we have a lot of duplicate packages. You can use DuplicatePackageCheckerPlugin and create a manuel resolution. They do that in banks https://github.com/cozy/cozy-banks/blob/f48951d8c4f89240da782611d1265b9f5a67bb24/config/webpack.config.manual-resolves.js#L28

That's true but I'm afraid of breaking a lot of things since duplicate packages are often of different major versions in cozy-notes

PR #205, opened 2021-08-11
staging: Fix resizing issues with media plugin

acezard:
Related to https://trello.com/c/OxuNeuMe/49-retour-2-lorsque-je-change-la-dimension-dune-image-non-centr%C3%A9e-je-ne-peux-plus-retourner-sur-sa-dimension-dorigine and https://trello.com/c/VmDB8tRA/48-retour-1-la-redimension-dune-image-non-centr%C3%A9e-nest-pas-prise-en-compte-lorsque-je-rafra%C3%AEchis

The issues were created by a mix of missing default values (that atlaskiteditor must somehow provide, but we don't have them here), and some other default values that were conflicting with our default image size (snap points to be precise). There was also a feature in mediaplugin that we didn't want, that was resizing aligned image to the smallest width possible. We wanted the opposite so modified the behaviour here.

The changelog is not updated because as the version is not yet released, it's technically not a bugfix, but part of building the 1.17 feature.


src/plugins/media/ui/ResizableMediaSingle/index.tsx
```diff
@@ -70,6 +71,9 @@ export default class ResizableMediaSingle extends React.Component<
     if (viewMediaClientConfig) {
       await this.checkVideoFile(viewMediaClientConfig)
     }
+
+    this.props.pctWidth &&
```
acezard, line 75: Without this, the component doesn't know the image has been resized and will always go back to the min width if we recenter the image. Strange behaviour but I figured this fix was simple enough.

Crash--: Maybe it's better to put that comment in the code itself?

acezard: Agreed

PR #206, opened 2021-08-12
feat: handle invalid file uploads

acezard:
Related to https://trello.com/c/y4QmVMRU/40-%F0%9F%93%9D-notes-affichage-derreur-si-upload-de-type-de-fichier-non-image

It is impossible to disallow the user to upload any kind of invalid file if he really wants to. The backend will error on such files. Thus, we will rely on it, there's no real gain to validate front-end side. That way we just wait to see if the upload failed with a 422, that will assure us the file is unprocessable. We just have to display an error alert for good UX as a way to handle this error.

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -89,7 +89,9 @@
     "unshared_text": "The owner may have revoked this sharing. You can no longer edit a note at this address.",
     "loading_error_title": "The note couldn't be loaded",
     "loading_error_text_noReturnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please try again later.",
-    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later."
+    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later.",
+    "422": "Only images are accepted",
+    "unknown_error": "An unexpected error occured"
   },
   "manifest": {
     "short_description": "Cozy Notes is your personal and collaborative note-taking application.",
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -1,0 +1,6 @@
+export interface CozyStackError {
+  status: number
+  title: string
+  detail: string
+  source: { parameter: string }
+}
--- a/src/components/notes/editor-view.jsx
+++ b/src/components/notes/editor-view.jsx
@@ -83,7 +83,7 @@ function EditorView(props) {
           appearance="full-page"
           placeholder={t('Notes.EditorView.main_placeholder')}
           shouldFocus={!readOnly}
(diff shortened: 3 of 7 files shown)


src/locales/en.json
```diff
@@ -89,7 +89,9 @@
     "unshared_text": "The owner may have revoked this sharing. You can no longer edit a note at this address.",
     "loading_error_title": "The note couldn't be loaded",
     "loading_error_text_noReturnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please try again later.",
-    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later."
+    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later.",
+    "422": "Only images are accepted",
```
acezard, line 93: @poupotte je te laisse valider les trads en/fr

src/locales/en.json
```diff
@@ -89,7 +89,9 @@
     "unshared_text": "The owner may have revoked this sharing. You can no longer edit a note at this address.",
     "loading_error_title": "The note couldn't be loaded",
     "loading_error_text_noReturnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please try again later.",
-    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later."
+    "loading_error_text_returnUrl": "The note may not exists anymore, or our servers may have a temporary glitch. Please go <a href='%{url}'>back</a> and try again later.",
```
Crash--, line 92: Is it used?

acezard: @Crash-- yes in `editor-loading-error.jsx`

src/types/errors.ts
```diff
@@ -0,0 +1,6 @@
+export interface CozyBackendError {
```
Crash--, line 1: Our vocabulary is about CozyStack, not CozyBackend.

src/types/errors.ts
```diff
@@ -0,0 +1,6 @@
...
+  status: number
+  title: string
+  detail: string
+  source: { parameter: string }
+}
```
Crash--, line 6: This type should be defined in CozyClient, not here.

src/locales/en.json
acezard: yea I was wondering 😅

Crash-- (approved):

src/types/errors.ts
acezard: Definitely as well as the Alerter methods. But in this context (prod monday) could we make an exception? I'll update the types later for cozy-client and cozy-ui

Crash--: The PR has been approved. So let's move on ;).

You can create an issue on cozy-client's side for the TypeError. 

For the Alerter, you can discuss about that here https://github.com/cozy/cozy-client/issues/552

PR #215, opened 2021-08-31
feat: Add cozy-to-cozy sharing

acezard:
Related to: https://trello.com/c/AiQg5Cjr/16-%F0%9F%93%9D-notes-partage-cozy-%C3%A0-cozy-notes-4j

Features:

- Add a new route to allow users to preview a note from email sharing
- Allows `Notes` to interact with `Contacts` (necessary for sharing to work as intended)
- Render the new `SharingBannerPlugin` in `Editor`; which manages itself
- Update the sharing modal in order to allow a user to share a note by email
- Update editor CSS overrides in order to display the `SharingBannerPlugin` in a `cozy-drive` way

Caveats:
- There are some questions about the post-sharing view. Should it go to the shared note or (currently) the shared note folder?
- Some old code about link sharing might be outdated by now. Should we review it entirely or continue to let it manage link sharing?

image

### DIFF
--- a/src/components/notes/editor-view.jsx
+++ b/src/components/notes/editor-view.jsx
@@ -8,6 +8,7 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import useEventListener from 'cozy-ui/transpiled/react/hooks/useEventListener'
 import Overlay from 'cozy-ui/transpiled/react/Overlay'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
+import { SharingBannerPlugin } from 'cozy-sharing'
 
 import editorConfig from 'components/notes/editor_config'
 import HeaderMenu from 'components/header_menu'
@@ -97,6 +98,7 @@ function EditorView(props) {
             t,
             setUploading
           )}
+          primaryToolbarComponents={<SharingBannerPlugin />}
           contentComponents={
             <WithEditorActions
               render={() => (
--- a/src/components/notes/editor-view.jsx
+++ b/src/components/notes/editor-view.jsx
@@ -8,7 +8,6 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import useEventListener from 'cozy-ui/transpiled/react/hooks/useEventListener'
 import Overlay from 'cozy-ui/transpiled/react/Overlay'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import { SharingBannerPlugin } from 'cozy-sharing'
 
 import editorConfig from 'components/notes/editor_config'
 import HeaderMenu from 'components/header_menu'
@@ -98,7 +97,7 @@ function EditorView(props) {
             t,
             setUploading
           )}
-          primaryToolbarComponents={<SharingBannerPlugin />}
+          primaryToolbarComponents={props.primaryToolbarComponents}
           contentComponents={
             <WithEditorActions
               render={() => (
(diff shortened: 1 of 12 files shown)


src/components/notes/sharing.jsx
```diff
@@ -7,27 +7,35 @@ import styles from 'components/notes/sharing.styl'
 
 export default function SharingWidget(props) {
   const client = useClient()
-
   const file = useFileWithPath({ cozyClient: client, file: props.file })
```
acezard, line 10: Why no file.name in hook return value?

src/styles/index.css
```diff
@@ -202,7 +202,7 @@ html .ProseMirror {
 }
 
 /* Titres */
-html body .note-editor-container .akEditor p,
+html body .note-editor-container .ak-editor-content-area p,
```
acezard, line 205: This isn't great but the banner will be broken otherwise. I don't think it's a problem though unless we intended to modify the styles of the editor toolbar paragraphs?

src/styles/index.css
```diff
@@ -252,7 +252,7 @@ html body .note-editor-container .akEditor > div:first-child  h6 {
 
 
 /* spacing between blocks */
-html body .note-editor-container .akEditor p,
+html body .note-editor-container .ak-editor-content-area p,
```
acezard, line 255: Same as above

src/styles/index.css
```diff
@@ -85,20 +85,21 @@ html .akEditor > div:first-child > div:first-child {
     overflow: visible; /* important for dropdowns */
 }
 
-/* avatars */
+/* second header block (contains sharing banner) */
```
acezard, line 88: This styling could be extracted in a proper stylus file but I thought it was more idiomatic to follow the initial convention of hacking editor-related things in this file. Thoughts?

src/components/notes/sharing.jsx
```diff
@@ -7,27 +7,35 @@ import styles from 'components/notes/sharing.styl'
...
+      <ShareButton
+        theme="primary"
+        docId={noteId}
+        onClick={onClick}
+        extension="narrow"
+        className={styles['sharing-button']}
+      />
+      {showModal && (
+        <ShareModal
+          document={{ ...file, name }}
```
acezard, line 33: cozy-client does not give us a filename here but it is absolutely needed in order to have a working sharing by email. Should we see why cozy-client does not give a filename at note root or is the current workaround good enough?

src/components/notes/editor-view.jsx
```diff
@@ -97,6 +98,7 @@ function EditorView(props) {
             t,
             setUploading
           )}
+          primaryToolbarComponents={<SharingBannerPlugin />}
```
acezard, line 101: Here we just inject the SharingBanner _inside_ the Atlaskit editor, as a custom component.

acezard: I have to see if the Editor can handle this props dynamically since Notes will manage the Banner display

src/components/notes/sharing.jsx
Crash--: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L125-L131

https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L111-L115


I don't see anything that remove the name attribute?

Crash-- (approved):

PR #223, opened 2021-09-22
feat: Improve list display in mobile view

acezard:
Related to: https://trello.com/c/bXVAxYj0/72-%F0%9F%93%9D-notes-mobile-afficher-le-chemin-des-notes-dans-la-liste-j https://trello.com/c/xTfcS4KC/73-lorsque-je-passe-sur-le-chemin-il-est-surlign%C3%A9-bleu-et-non-noir

Request: show note name in mobile

Implementation: create a WithBreakpoints component and improve responsive handling in List

image

### DIFF
--- a/src/components/notes/List/List.jsx
+++ b/src/components/notes/List/List.jsx
@@ -1,4 +1,5 @@
 import React from 'react'
+
 import {
   Table,
   TableHead,
@@ -8,13 +9,14 @@ import {
   TableCell
 } from 'cozy-ui/transpiled/react/Table'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
-
 import { translate } from 'cozy-ui/transpiled/react/I18n'
+
 import EmptyComponent from 'components/notes/List/EmptyComponent'
 import NoteRow from 'components/notes/List/NoteRow'
-
 import styles from 'components/notes/List/list.styl'
+import { WithBreakpoints } from './WithBreakpoints'
+import { Breakpoints } from 'types/enums'
+
 const EmptyTableRow = () => (
   <TableRow className={styles.tableSpecialRow}>
     <TableCell className={styles.tableCellEmpty}>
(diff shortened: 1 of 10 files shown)


src/components/notes/List/List.jsx
```diff
@@ -8,13 +9,14 @@ import {
...
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
-
 import { translate } from 'cozy-ui/transpiled/react/I18n'
+
 import EmptyComponent from 'components/notes/List/EmptyComponent'
 import NoteRow from 'components/notes/List/NoteRow'
-
 import styles from 'components/notes/List/list.styl'
+import { WithBreakpoints } from './WithBreakpoints'
```
Crash--, line 17: Can't you use one of the already existing class? https://docs.cozy.io/cozy-ui/styleguide/section-utilities.html#kssref-utilities-hide

Crash--: Also, I'm not a big fan of adding this kind of HOC. What is the gain compared to 

```
{!isMobile && <TableHeader... }
```

There is no API creation, no new component, no new HOC.

acezard: Point 1: I don't want to use CSS classes here because I want to avoid a render, not hide Elements in the DOM

Point 2: While it's inspired by the HoC pattern I wouldn't say it's a HoC, more like a wrapper component (it's a component itself, not a function taking a component as argument and returning a component). 

What's the advantage compared to `isMobile`? 

First, it allows me to use the hook `isMobile`  a lot less than before everywhere in `cozy-notes`.

Secondly, it allows me to use a declarative style in my view components instead of an imperative one, which while subjective, looks clearer to me and removes logical operators from the view.

Finally, it allows me, with the help of typings, to have more control and knowledge about which breakpoints I can use to hide or show something depending on the design requirements.

I don't think I'm creating anything heavy here, just types (deducted from existing code) and a very simple wrapper component.

Crash--: Let's add this HOC in Cozy-UI then (you can still keep it here, write a todo somewhere saying use the one from ui once this PR XXX get released for instance)

Crash-- (approved):

src/components/notes/List/List.jsx
acezard: Yes that was my idea but it's very important to have type declarations here. We can't allow incorrect values in `showOn` or  `hideOn` and the component must provide explicitly what values it will accept. Cozy-ui doesn't have declaration files at the moment (nor is it configured to allow TS files, but it could be), do you have a suggestion on how to handle that issue?

acezard: After thinking about it I guess I could create a handwritten declaration file for cozy-ui (that is, not generated by jsdoc)

PR #224, opened 2021-09-22
feat: Remove filename

acezard:
Related to: https://trello.com/c/wozfDjxO/75-%F0%9F%93%9D-notes-ne-pas-afficher-le-nom-dans-le-chemin-de-la-note-mais-que-son-chemin

Simply display the path of a note without `/{noteName}`  by using a small RegExp.exec() call

image

### DIFF
--- a/src/components/notes/List/NotePath.tsx
+++ b/src/components/notes/List/NotePath.tsx
@@ -4,7 +4,7 @@ import FilePathLink from 'cozy-ui/transpiled/react/FilePathLink'
 import AppLinker from 'cozy-ui/transpiled/react/AppLinker'
 
 import { Slugs } from 'constants/strings'
-import { stopPropagation } from 'lib/helpers'
+import { removeFilename, stopPropagation } from 'lib/helpers'
 
 export const NotePath: React.FC<{
   drivePath: string
@@ -20,7 +20,7 @@ export const NotePath: React.FC<{
           ? { target: '_blank', rel: 'noreferrer noopener' }
           : { target })}
       >
-        {path}
+        {removeFilename(path)}
       </FilePathLink>
     )}
   </AppLinker>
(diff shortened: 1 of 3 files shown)


src/components/notes/List/NotePath.tsx
```diff
@@ -20,7 +20,7 @@ export const NotePath: React.FC<{
           ? { target: '_blank', rel: 'noreferrer noopener' }
           : { target })}
       >
-        {path}
+        {/(.+)(\/.+)$/.exec(path)?.[1]}
```
JF-Cozy, line 23: Est-ce qu'on peut ajouter un test ou deux pour être sûr stp ?
Est-ce qu'on n'a pas déjà l'info dans un autre object ?

acezard: J'ai cherché mais pas trouvé. Pas trouvé non plus quelquechose du genre splitFileName. Je peux ajouter des tests oui

acezard: tests ajoutés

src/lib/helpers.ts
```diff
@@ -4,3 +4,6 @@ export const processFile = (
   file === null || typeof file === 'string' ? false : file
 
 export const stopPropagation = (event: Event): void => event.stopPropagation()
+
+export const removeFilename = (path: string): string =>
+  /(.+)(\/.+)$/.exec(path)?.[1] || path
```
Crash--, line 9: I would have done something like this: 
- we have the filename 
- we have the fullpath containing the filename

I'd have only do something like filePath.replace(`${filename}$`, '') no? It seems easier to read and understand.

Crash-- (approved):

src/lib/helpers.ts
acezard: Yeah, that also works but I wanted to keep props/args at a minimum level. In terms of performance, I'm not sure if there's a big difference between the two approaches. But in terms of API, I think it's nice to have a single arg function.

PR #227, opened 2021-09-24
feat: Add drag and drop for images

acezard:
### Implementation issues

- Atlaskit editor does not handle by default this feature
- It was impossible to activate it, nowhere in the codebase is it mentioned
- ProseMirror does have a plugin handling it, but Atlaskit does not use it

### Implementation

- Externalize `image-upload` plugin from Atlaskit to Cozy
- Add a function in that plugin that will set the editor selection to the drop location
- Improve image uploader to handle drop events
- Disallow the user to drop images outside ProseMirror with a global listener

### DIFF
--- a/src/lib/image-upload-provider.ts
+++ b/src/lib/image-upload-provider.ts
@@ -36,8 +36,14 @@ export const imageUploadProvider = (
   )
 
 const handleEvent = (event: Event | undefined, cb: FileCallback): void => {
-  if (event === undefined) return handleClick(cb)
-  if (event.type === EventType.Drop) return handleDrop(event as DragEvent, cb)
+  switch (event?.type) {
+    case EventType.Drop:
+      return handleDrop(event as DragEvent, cb)
+    case EventType.Paste:
+      return handlePaste(event as ClipboardEvent, cb)
+    default:
+      return handleClick(cb)
+  }
 }
 
 const handleClick = (cb: FileCallback): void => {
@@ -56,6 +62,9 @@ const handleClick = (cb: FileCallback): void => {
 
 const handleDrop = (event: DragEvent, cb: FileCallback): void =>
   cb(event.dataTransfer?.files?.[0])
+
+const handlePaste = (event: ClipboardEvent, cb: FileCallback): void =>
+  cb(event.clipboardData?.files?.[0])
 
 const uploadImage = (
   config: ImageUploadProviderConfig,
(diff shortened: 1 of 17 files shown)


src/lib/image-upload-provider.ts
```diff
@@ -19,58 +20,83 @@ interface CollabProvider {
...
+
+  inputElement.dispatchEvent(new MouseEvent(EventType.Click))
+
+  inputElement.addEventListener(EventType.Change, () =>
+    cb(inputElement.files?.[0])
+  )
+}
+
+const handleDrop = (event: DragEvent, cb: FileCallback): void =>
+  cb(event.dataTransfer?.files?.[0])
```
Crash--, line 64: Should not you check that the files[0] is really a file (and not a folder) and specially an image?

src/plugins/image-upload/event-handlers.ts
```diff
@@ -0,0 +1,11 @@
...
+  'dragover',
+  event => (
+    event.preventDefault(),
+    !(<HTMLElement>event?.target).closest('.ProseMirror') &&
+      event.dataTransfer !== null &&
+      (event.dataTransfer.dropEffect = 'none')
+  )
+)
+
+export {}
```
Crash--, line 11: No removeEventListener somewhere?

src/lib/image-upload-provider.ts
```diff
@@ -57,6 +63,9 @@ const handleClick = (cb: FileCallback): void => {
 const handleDrop = (event: DragEvent, cb: FileCallback): void =>
   cb(event.dataTransfer?.files?.[0])
 
+const handlePaste = (event: ClipboardEvent, cb: FileCallback): void =>
+  cb(event.clipboardData?.files?.[0])
```
Crash--, line 67: Same, no check on the file?

Crash-- (approved):
A small question but overall looks good! 👏

src/lib/image-upload-provider.ts
acezard: Well, I'm letting the back-end do that, figured it would be simpler as we have a toaster in case the file is not accepted

src/plugins/image-upload/event-handlers.ts
acezard: Yes, I decided it was not really worth it because the listener will always be needed unless the user leaves the application.

src/lib/image-upload-provider.ts
acezard: yes the back-end will refuse a non-image (or directory) and a toaster will appear

PR #236, opened 2021-10-15
feat: Implement MVP header

acezard:
image

J'avais travaillé depuis le début en essayant de convertir la toolbar de drive en toolbar agnostique dans cozy-ui, malheureusement plus j'avançais plus j'avais l'impression de créer de la dette technique et de créer un composant très complexe avec beaucoup trop de props pour qu'il soit compatible à la fois avec drive et notes.

Etant insatisfait avec cela j'ai décidé d'annuler le travail qui était fait sur cozy-ui depuis la toolbar cozy-drive et j'ai recommencé from scratch sur cozy-notes, en utilisant cependant uniquement des composants déjà existants. Je n'ai malheureusement pas eu le temps de terminer cette approche en raison du switch trop tardif. Cela-dit, étant donné que les composants utilisés existent déjà, le MVP est fonctionnel.

En terme de reste à faire, il y a une anomalie sur le path de la note qui n'est pas fetch au bon endroit, et il faut aussi vérifier que le bouton de retour répond bien à la spec. Il y a aussi des possibilités de refactor. Quoi qu'il en soit en restant sur cette base découplée de la toolbar cozy-drive je pense que ces 3 points peuvent être terminés en un jour.

### DIFF
--- a/src/components/header_menu.jsx
+++ b/src/components/header_menu.jsx
@@ -1,11 +1,81 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 
-const HeaderMenu = props => (
-  <header className={'page-header-menu ' + (props.className || '')}>
-    <div className="page-header-menu--left">{props.left}</div>
-    {props.children}
-    <div className="page-header-menu--right">{props.right}</div>
-  </header>
-)
+import AppLinker from 'cozy-ui/transpiled/react/AppLinker'
+import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
+import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider'
+import Link from 'cozy-ui/transpiled/react/Link'
+import { SharedRecipients } from 'cozy-sharing'
+import { Typography } from '@material-ui/core'
+import { Slugs } from 'constants/strings'
+import { NotePath } from './notes/List/NotePath'
+import { CozyFile } from 'cozy-doctypes'
+import { getDriveLink } from 'lib/utils'
+
+import styles from './header_menu.styl'
+import { WithBreakpoints } from './notes/List/WithBreakpoints'
+import { Breakpoints } from 'types/enums'
+
+const HeaderMenu = ({
+  homeHref,
+  backFromEditing,
+  editorCorner,
+  isPublic,
+  file,
+  client
+}) => {
+  const drivePath = useMemo(
+    () => getDriveLink(client, file.attributes.dir_id),
+    [client, file.attributes.dir_id]
+  )
+
+  const { filename } = CozyFile.splitFilename(file.attributes)
+
+  return (
+    <header className={styles['header-menu']}>
+      <WithBreakpoints hideOn={Breakpoints.Mobile}>
+        <AppLinker slug={Slugs.Home} href={homeHref}>
+          {({ href }) => (
+            <Link {...(!isPublic && { href })} className={styles['home-link']}>
+              <AppIcon app={Slugs.Home} alt={Slugs.Home} />
+            </Link>
+          )}
+        </AppLinker>
+
+        <Divider orientation="vertical" className={styles['divider']} />
+      </WithBreakpoints>
+
+      {backFromEditing && <div className="u-mr-1">{backFromEditing}</div>}
+
+      <div className={styles['file-infos']}>
+        <WithBreakpoints hideOn={Breakpoints.Mobile}>
+          <AppIcon
+            app={file.attributes.cozyMetadata.createdByApp}
+            className={styles['app-icon']}
+          />
+        </WithBreakpoints>
+
+        <div>
+          <Typography>
+            <strong>{filename}</strong>
+          </Typography>
+
+          <WithBreakpoints hideOn={Breakpoints.Mobile}>
+            <NotePath
+              drivePath={drivePath}
+              path={file.attributes.path || drivePath}
+              target="_blank"
+            />
+          </WithBreakpoints>
+        </div>
+      </div>
+
+      <WithBreakpoints hideOn={Breakpoints.Mobile}>
+        <SharedRecipients docId={file.id} size={24} />
+      </WithBreakpoints>
+
+      {editorCorner}
+    </header>
+  )
+}
 
 export default HeaderMenu
--- a/src/components/header_menu.jsx
+++ b/src/components/header_menu.jsx
@@ -21,7 +21,8 @@ const HeaderMenu = ({
   editorCorner,
   isPublic,
   file,
-  client
+  client,
+  primaryToolBarComponents
 }) => {
   const drivePath = useMemo(
     () => getDriveLink(client, file.attributes.dir_id),
@@ -62,7 +63,7 @@ const HeaderMenu = ({
           <WithBreakpoints hideOn={Breakpoints.Mobile}>
             <NotePath
               drivePath={drivePath}
-              path={file.attributes.path || drivePath}
+              path={file.attributes.path || drivePath.split('#')[1]}
               target="_blank"
             />
           </WithBreakpoints>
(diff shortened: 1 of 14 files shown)


src/components/header_menu.jsx
```diff
@@ -1,11 +1,81 @@
...
-  </header>
-)
+import React, { useMemo } from 'react'
+
+import AppLinker from 'cozy-ui/transpiled/react/AppLinker'
+import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
+import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider'
+import Link from 'cozy-ui/transpiled/react/Link'
+import { SharedRecipients } from 'cozy-sharing'
+import { Typography } from '@material-ui/core'
```
Crash--, line 8: You have to use Typography from cozy-ui, not from mui.

src/components/header_menu.jsx
```diff
@@ -1,11 +1,81 @@
...
+import { Slugs } from 'constants/strings'
+import { NotePath } from './notes/List/NotePath'
+import { CozyFile } from 'cozy-doctypes'
```
Crash--, line 11: To manipulate files, you should not use cozy-doctypes anymore. All the methods have been moved to cozy-client : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L24

src/components/header_menu.jsx
```diff
@@ -1,11 +1,81 @@
...
+import { getDriveLink } from 'lib/utils'
+
+import styles from './header_menu.styl'
+import { WithBreakpoints } from './notes/List/WithBreakpoints'
+import { Breakpoints } from 'types/enums'
```
Crash--, line 16: Please respect import orders: 
1/ third parties
2/ cozy's lib 
3/ internal import

src/components/header_menu.jsx
```diff
@@ -1,11 +1,81 @@
...
+}) => {
+  const drivePath = useMemo(
+    () => getDriveLink(client, file.attributes.dir_id),
+    [client, file.attributes.dir_id]
+  )
+
+  const { filename } = CozyFile.splitFilename(file.attributes)
+
+  return (
+    <header className={styles['header-menu']}>
```
Crash--, line 35: We should avoid creating custom css. Can't we use our utilities? https://docs.cozy.io/cozy-ui/styleguide/section-utilities.html

src/components/header_menu.jsx
```diff
@@ -1,11 +1,81 @@
...
+      <WithBreakpoints hideOn={Breakpoints.Mobile}>
+        <AppLinker slug={Slugs.Home} href={homeHref}>
+          {({ href }) => (
+            <Link {...(!isPublic && { href })} className={styles['home-link']}>
+              <AppIcon app={Slugs.Home} alt={Slugs.Home} />
+            </Link>
+          )}
+        </AppLinker>
+
+        <Divider orientation="vertical" className={styles['divider']} />
```
Crash--, line 45: same here, use our margin's utilities

src/styles/index.css
```diff
@@ -77,10 +77,19 @@ html .akEditor > div:first-child {
...
 /* prose mirror toolbar in the upper bar */
 html .akEditor > div:first-child {
-    margin-bottom: 0;
+    width: 100%;
     flex-wrap: wrap;
     height: auto;
-    box-shadow: 0 11px 9px 0 rgba(0, 0, 0, 0.08), 0 3px 3px 0 rgba(50, 54, 63, 0.08);
+    position: fixed;
+    top: auto;
+    bottom: 1rem;
```
Crash--, line 85: What about the `safe-area` for mobile browser? Should we respect it?

src/components/header_menu.jsx
```diff
@@ -62,18 +63,20 @@ const HeaderMenu = ({
           <WithBreakpoints hideOn={Breakpoints.Mobile}>
             <NotePath
               drivePath={drivePath}
-              path={file.attributes.path || drivePath}
+              path={file.attributes.path || drivePath.split('#')[1]}
```
Crash--, line 66: why this change?

src/components/header_menu.styl
```diff
@@ -1,8 +1,11 @@
 .header-menu
     align-items center
+    box-shadow 0px 0px 0px 0.5px rgba(29, 33, 42, 0.12), 0px 2px 4px rgba(29, 33, 42, 0.0793047), 0px 4px 16px rgba(29, 33, 42, 0.06)
     display flex
     justify-content space-between
     padding 0.5rem 1rem
+    position relative
+    z-index 1
```
Crash--, line 8: why this z-index? What is the purpose? You should use our css's var to handle zIndex https://github.com/cozy/cozy-ui/blob/67485ca3a6db28dd4ba64d2c91188c21b00034ce/stylus/settings/z-index.styl#L55-L69

src/components/header_menu.jsx
acezard: Will do

src/components/header_menu.jsx
acezard: Wasn't aware of that, thanks. Will do

src/components/header_menu.jsx
acezard: Oversight from my end, usually I respect the order. Will update

src/components/header_menu.jsx
acezard: I am going to dig into that but I can't guarantee I will find everything I need to avoid CSS

src/components/header_menu.jsx
acezard: Will do

src/styles/index.css
acezard: I am sorry I am not sure what you're referring to in this context, can you please elaborate?

src/components/header_menu.jsx
acezard: It allows me to have the drive path without querying the actual drive directory. I found it to be pretty effective since it eliminates one http request

src/components/header_menu.styl
acezard: Will use a variable. The z-index is necessary for the header drop shadow to spread over the actual editor area (otherwise the editor area will erase the shadow).

src/components/header_menu.jsx
acezard: After looking into it there's too much CSS I don't think I can translate it in atomic classes. Probably would be best to make a cozy-ui component (this toolbar is used by 3 apps). Should I create an issue?

src/components/header_menu.jsx
Crash--: OK why not, but I still don't understand why we need to use the drivePath for that. I don't understand why we need this `|| drivePath`? Does it mean that `file.attributes.path` can be empty? 🤔 

I don't think that relying on an internal stuff like `#` is a good way to handle it if needed. Right now you use `generateWebLink` but pretty soon we'll use an other method and the `#` will not be there anymore. This is something to avoid.

src/styles/index.css
Crash--: On mobile we've a concept called "safe area" that refers to the space where... it's safe to display content. For exemple  iPhone X and + have a "gray bar" that you can drag to go back to the home screen. Since this "gray bar" is interactive it's not safe to display content close to it. (https://developer.apple.com/documentation/uikit/uiview/positioning_content_relative_to_the_safe_area) 

There are few CSS's vars to handle that correctly. You can take a look at our CozyDialog implementation in Cozy UI if you want. (for instance https://github.com/cozy/cozy-ui/pull/1677) 

Since this code seems to have a fixed position at bottom: 1rem I'm asking if we should not take care of this safe are

acezard: I suggest  `bottom: calc(env(safe-area-inset-bottom, 1rem) + 1rem);`

Crash-- (approved):

PR #243, opened 2021-12-16
fix: Properly display note path

acezard:
Related to https://trello.com/c/wpOZ52TO/95-%F0%9F%93%9D-notes-ne-pas-afficher-le-chemin-complet-dune-note-partag%C3%A9e

Using useFileWithPath() hook to get huiman readable filepath name

### DIFF
--- a/src/components/header_menu.jsx
+++ b/src/components/header_menu.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React from 'react'
 
 import { models } from 'cozy-client'
 import { SharedRecipients } from 'cozy-sharing'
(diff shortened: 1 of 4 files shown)


src/components/header_menu.jsx
```diff
@@ -14,23 +14,23 @@ import { getDriveLink } from 'lib/utils'
...
+import useFileWithPath from 'hooks/useFileWithPath'
 
 const HeaderMenu = ({
   homeHref,
   backFromEditing,
   editorCorner,
   isPublic,
   file,
-  client,
+  cozyClient,
```
Crash--, line 25: to me you should not pass the client in props. You should use `useClient()` inside the component to get it

acezard: updated

src/components/header_menu.jsx
```diff
@@ -61,13 +61,15 @@ const HeaderMenu = ({
...
-          </WithBreakpoints>
+          {fileWithPath?.path && (
+            <WithBreakpoints hideOn={Breakpoints.Mobile}>
+              <NotePath
+                drivePath={drivePath}
+                path={fileWithPath.path}
+                target="_blank"
+              />
+            </WithBreakpoints>
+          )}
```
Crash--, line 72: I don't know if this is the right fix. 

What are you displaying if you share by link a note?
What are you displaying if you access a not inside a sub sub shared folder?

acezard:
use le hook getpathname dans un nouveau component si is not public

Crash-- (approved):

PR #245, opened 2021-12-16
fix: display icons in public mode

acezard:
Show static assets for unknown users

### DIFF
--- a/src/components/header_menu.jsx
+++ b/src/components/header_menu.jsx
@@ -14,6 +14,7 @@ import { getDriveLink } from 'lib/utils'
 import { NotePath } from './notes/List/NotePath'
 import { WithBreakpoints } from './notes/List/WithBreakpoints'
 import styles from './header_menu.styl'
+import { useFetchIcons } from 'hooks/useFetchIcons'
 
 const HeaderMenu = ({
   homeHref,
@@ -29,7 +30,7 @@ const HeaderMenu = ({
     [client, file.attributes.dir_id]
   )
   const simplifiedDrivePath = drivePath.split('#')[1]
-
+  const { fetchHomeIcon, fetchNoteIcon } = useFetchIcons()
   const { filename } = models.file.splitFilename(file.attributes)
 
   return (
@@ -38,7 +39,11 @@ const HeaderMenu = ({
         <AppLinker slug={Slugs.Home} href={homeHref}>
           {({ href }) => (
             <Link {...(!isPublic && { href })} className={styles['home-link']}>
-              <AppIcon app={Slugs.Home} alt={Slugs.Home} />
+              <AppIcon
+                app={Slugs.Home}
+                alt={Slugs.Home}
+                fetchIcon={fetchHomeIcon}
+              />
             </Link>
           )}
         </AppLinker>
@@ -53,6 +58,7 @@ const HeaderMenu = ({
           <AppIcon
             app={file.attributes.cozyMetadata.createdByApp}
             className={styles['app-icon']}
+            fetchIcon={fetchNoteIcon}
           />
         </WithBreakpoints>
 
--- a/src/hooks/useFetchIcons.ts
+++ b/src/hooks/useFetchIcons.ts
@@ -1,0 +1,19 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+import { useClient } from 'cozy-client'
+
+export const useFetchIcons = (): {
+  fetchHomeIcon: () => string
+  fetchNoteIcon: () => string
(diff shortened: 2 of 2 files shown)


src/components/header_menu.jsx
```diff
@@ -35,13 +35,17 @@ const HeaderMenu = ({
...
     <header className={styles['header-menu']}>
       <WithBreakpoints hideOn={Breakpoints.Mobile}>
-        <AppLinker slug={Slugs.Home} href={homeHref}>
-          {({ href }) => (
-            <Link {...(!isPublic && { href })} className={styles['home-link']}>
-              <AppIcon app={Slugs.Home} alt={Slugs.Home} />
-            </Link>
-          )}
-        </AppLinker>
+        {homeHref ? (
```
acezard, line 38: homeHref is always defined even with a 0 cookie user.  
`collabProvider.serviceClient.cozyClient.getStackClient().uri`

Is there an easy way to differentiate a cozy user previewing the page from a non cozy user? The token in the url makes real cozy users and non cozy users looking the same in the client

src/components/IconCozyHome.jsx
```diff
@@ -0,0 +1,33 @@
...
+    `${client.get.cozyURL()}/assets/images/icon-cozy-home.svg`
+
+  return (
+    <AppIcon
+      fetchIcon={fetchIcon}
+      fallbackIcon={SvgIconCozyHome}
+      className={className}
+    />
+  )
+}
```
Crash--, line 33: Can't we use `Cozy` icon from UI? 🤔 https://docs.cozy.io/cozy-ui/react/#!/Icon/5

src/components/header_menu.jsx
Crash--: Make a search on `isPublic` into the codebase, you'll find what you want, I think. You removed it in this component...

src/components/header_menu.styl
```diff
@@ -10,8 +10,8 @@
 .app-icon
 .home-link
     display block
-    height 32px
-    width 32px
+    height 32px !important
+    width 32px !important
```
Crash--, line 14: why?

src/components/IconCozyHome.jsx
```diff
@@ -0,0 +1,33 @@
...
+        />
+      </g>
+    </svg>
+  )
+}
+
+export const IconCozyHome = ({ className }) => {
+  const client = useClient()
+  const fetchIcon = () =>
+    `${client.get.cozyURL()}/assets/images/icon-cozy-home.svg`
```
Crash--, line 24: This since icon should be served for our public page, it should be placed under a /public/ folder and the manifest should be updated... 

Also, what is this get.cozyURL() stuff? I don't see anything like this in cozyclient ? 🤔

src/components/header_menu.jsx
Crash--: I don't understand the meaning of `homeHref`. We should remove this concept and restore `isPublic` stuff.

Also, we should only use the SVG from the public folder for the both case. The only stuff that is changing is that we don't put the href to the link if we're in public. Like we did before.

Crash-- (changes requested):
See comment

src/components/header_menu.jsx
```diff
@@ -35,13 +35,17 @@ const HeaderMenu = ({
   return (
     <header className={styles['header-menu']}>
       <WithBreakpoints hideOn={Breakpoints.Mobile}>
-        <AppLinker slug={Slugs.Home} href={homeHref}>
-          {({ href }) => (
-            <Link {...(!isPublic && { href })} className={styles['home-link']}>
-              <AppIcon app={Slugs.Home} alt={Slugs.Home} />
```
acezard, line 41: fetchIcon props `${client.get.cozyURL()}/assets/images/icon-cozy-home.svg`

src/components/header_menu.jsx
```diff
@@ -50,10 +54,15 @@ const HeaderMenu = ({
 
       <div className={styles['file-infos']}>
         <WithBreakpoints hideOn={Breakpoints.Mobile}>
-          <AppIcon
-            app={file.attributes.cozyMetadata.createdByApp}
-            className={styles['app-icon']}
-          />
+          {homeHref ? (
```
acezard, line 57: isPublic

src/components/header_menu.jsx
```diff
@@ -50,10 +54,15 @@ const HeaderMenu = ({
...
+            <AppIcon app={Slugs.Notes} className={styles['app-icon']} />
```
acezard, line 58: fetchIcon props route publique

Crash-- (approved):
nit: I don't like the use of `hook` folder neither the name of `fetch` since we do not make API request but...

PR #246, opened 2021-12-16
fix: Hide empty bar under toolbar

acezard:


src/styles/index.css
```diff
@@ -131,6 +131,11 @@ html .akEditor > div:first-child > div:first-child + div hr {
     height: calc(100%);
 }
 
+/* Hacky CSS to hide anything that isn't the toolbar in the atlaskit toolbar wrapper (avatar component for instance) */
+[data-testid="ak-editor-main-toolbar"] > :not(:first-child) {
+    display: none !important;
+}
+
```
Crash--, line 138: I'm not confortable with this fix. 

Just below we already have CSS that hides avatar. Can't we fix it? 
https://github.com/cozy/cozy-notes/blob/aa444ff6c169a8c3c311cb508f917bb39a88bbd5/src/styles/index.css#L139-L146

acezard: @Crash-- no, the fixme you're linking only hide the avatar, but not their wrapper. I'm hiding their wrapper with a wildcard (heavy handed sure, but there should only be the editor toolbar in there)

Crash-- (approved):
🤷‍♂️

acezard:
remove old fixme

PR #247, opened 2021-12-17
fix: Allow tabbing out of the title

acezard:
https://trello.com/c/cz5BooES/98-%F0%9F%93%9D-notes-pouvoir-sortir-du-titre-apr%C3%A8s-un-entr%C3%A9e

### DIFF
--- a/src/hooks/useTitleChanges.js
+++ b/src/hooks/useTitleChanges.js
@@ -6,6 +6,12 @@ function useTitleChanges({ noteId, title, setTitle, serviceClient }) {
   const onLocalTitleChange = useCallback(
     serviceClient
       ? e => {
+          // Let the user tab out of the title when using the 'Enter' key
+          if (e.nativeEvent?.inputType === 'insertLineBreak')
+            return window.dispatchEvent(
+              new KeyboardEvent('keydown', { key: 9 })
+            )
+
           // forbid line feed and non standard spaces in title
           const modifiedTitle = e.target.value.replace(/[\r\t\n\xa0]+/g, ' ')
           if (title != modifiedTitle) {


Crash-- (changes requested):
see comment

acezard:
> `with-outer-listeners` is only used on `DropDown` stuff. We can't rely on this to provide this kind of side effect. If one day we decide to remove this DropDown UI, we'll break this kind of stuff.
> 
> We already have logic to handle the title attribute. We should add this logic there https://github.com/cozy/cozy-notes/blob/master/src/components/notes/editor-view.jsx#L64-L66

I agree my implementation is most hacky. But we can't use onTitleBlur, we never trigger blur in the current implementation. Please have a call to talk about this

Crash--:
Let's create a new listener for the textarea. But not in `with-outer-listeners`. We should add it to `EditorView`

acezard:
useTitleChanges

src/hooks/useTitleChanges.js
```diff
@@ -6,6 +6,10 @@ function useTitleChanges({ noteId, title, setTitle, serviceClient }) {
   const onLocalTitleChange = useCallback(
     serviceClient
       ? e => {
+          // Let the user tab out of the title when using the 'Enter' key
+          if (e.nativeEvent?.inputType === 'insertLineBreak')
+            return window.document.querySelector('.ProseMirror').focus()
```
Crash--, line 11: This is not what we asked. We want a "tab", not a focus on this selector

acezard: But there is no API to ask a tab event to the browser. The best we can do is find the next focusable element (tabindex), but some focusable elements do not have tabindex. What would you suggest?

Crash--: https://developer.mozilla.org/fr/docs/Web/API/EventTarget/dispatchEvent
```
window.dispatchEvent(new KeyboardEvent('keydown', {
  'key': 9
}));
```

will work, no?

acezard: It does, that's smart

acezard: @Crash-- updated

Crash-- (approved):

PR #250, opened 2021-12-20
fix: Update floating toolbar zindex

acezard:
https://trello.com/c/GGMuFgYs/106-%F0%9F%93%9D-notes-bugs-barre-doutils-image

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -453,7 +453,7 @@ html .fabric-editor-popup-scroll-parent > div > div {
 [aria-label="Media floating controls"],
 .richMedia-resize-handle-left,
 .richMedia-resize-handle-right {
-    z-index: auto !important;
+    z-index: 1 !important;
 }
 
 /* Required for toolbar's popups because unidentified internals miscalculate the popups position */


Crash-- (changes requested):
We want from 60 to auto in order to mitigate a bug with toolbar https://github.com/cozy/cozy-notes/commit/6e152a842b16f50e8d5e147e6f5cb284948c2a97#diff-c1c322616098d4351e10768cd3ad654e2b79fcf30cff1606ff4e79b7893fbe5bR440

Now we're using 1... I don't understand our strategy here. Can you, at least, explain in the commit why 1 is working?

acezard:
> We want from 60 to auto in order to mitigate a bug with toolbar [6e152a8#diff-c1c322616098d4351e10768cd3ad654e2b79fcf30cff1606ff4e79b7893fbe5bR440](https://github.com/cozy/cozy-notes/commit/6e152a842b16f50e8d5e147e6f5cb284948c2a97#diff-c1c322616098d4351e10768cd3ad654e2b79fcf30cff1606ff4e79b7893fbe5bR440)
> 
> Now we're using 1... I don't understand our strategy here. Can you, at least, explain in the commit why 1 is working?

Well it's above the various note components. And the lowest level possible, so below cozy components. I admit the z-index issues have been numerous. Too many clashes between different systems.

Crash-- (approved):

PR #255, opened 2022-01-03
fix: Allow tabbing out of the title

acezard:
Dispatchevent did not work


Crash--:
On which browser? Didn't you test before making the change?

acezard:
> On which browser? Didn't you test before making the change?

Yes I tested and it worked at the time, I must have tested wrong. Browsers do not allow to trigger a real-life tab event for security reasons from what I understood (can't find the source at the moment)

Crash-- (approved):

PR #287, opened 2022-05-03
fix: Sharing note with empty title

Crash--:
If the user didn't write any title,
then the props.title is undefined
In that case, we fallback to the
file.attributes.names. It'll be something
like "New Note ...."

Cozy-Stack doesn't allow us to create
a sharing's rule with an empty string
as title.


acezard (approved):

trollepierre (approved):

nono:
Ping @Crash--

PR #288, opened 2022-05-14
fix: Stay in App Amirale when opening a Note

acezard:
https://trello.com/c/Ntqibd4M/428-%F0%9F%90%9B-lapp-notes-mouvre-drive-dans-le-navigateur-plut%C3%B4t-que-lapp-amirale

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "cozy-harvest-lib": "7.2.3",
     "cozy-intent": "^1.17.1",
     "cozy-realtime": "4.0.5",
-    "cozy-scripts": "6.1.1",
+    "cozy-scripts": "^6.3.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
     "cozy-harvest-lib": "7.2.3",
     "cozy-intent": "^1.17.1",
     "cozy-realtime": "4.0.5",
-    "cozy-scripts": "6.1.1",
+    "cozy-scripts": "^6.3.1",
```
Crash--, line 68: This is not a chore commit. cozy-scripts can inject / modify bundled code.

Also did you test the public view? Is it still working?

Crash-- (approved):
👍

PR #293, opened 2022-06-27
fix: Update cozy-client

acezard:
This should fix HTTPS issues happening now in iOS,
when coming from another application.


Ldoppea:
Shouldn't we add `@cozy/minilog` and `cozy-logger` in dependencies as described in the changelog? https://github.com/cozy/cozy-client/blob/master/CHANGELOG.md#2800-2022-04-08

acezard:
> Shouldn't we add `@cozy/minilog` and `cozy-logger` in dependencies as described in the cozy-client's changelog? https://github.com/cozy/cozy-client/blob/master/CHANGELOG.md#2800-2022-04-08

@Ldoppea added

Ldoppea (approved):

PR #294, opened 2022-06-28
chore: Upgrade ubuntu to focal

Crash--:
xenial is still the default ubuntu version
on Travis, but do not have standart support
anymore.

So let's use the latest available LTS at this
time.

See https://wiki.ubuntu.com/Releases for
more informations about ubuntu support.


sblaisot:
latest... available 👌 
😉

PR #298, opened 2022-07-15
feat: Upgrade react-router-dom to v6.3.0

acezard:
* Use <Route children> instead of <Route render> and/or <Route component> props

* Use our hooks API to access router state like the current location and params

* Replace all uses of withRouter with hooks

* Replace any <Route>s that are not inside a <Switch> with useRouteMatch, or wrap them in a <Switch>

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -1,5 +1,4 @@
-/* global cozy */
-import React, { useState, useEffect, useMemo } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import {
   HashRouter,
   Route,
@@ -24,12 +23,12 @@ import { useClient } from 'cozy-client'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
 const manifest = require('../../manifest.webapp')
+import { AppRoutes } from 'constants/routes'
 import { List, Editor, Unshared } from 'components/notes'
+import { fetchIfIsNoteReadOnly } from 'lib/utils'
+import { getDataOrDefault } from 'lib/initFromDom'
 import { getReturnUrl, getSharedDocument } from 'lib/utils'
 import { useFlagSwitcher } from 'lib/debug'
-import { getDataOrDefault } from 'lib/initFromDom'
-import { fetchIfIsNoteReadOnly } from 'lib/utils'
-import { AppRoutes } from 'constants/routes'
 
 const RoutedEditor = () => {
   const { id } = useParams()
@@ -51,11 +50,16 @@ const PrivateContext = () => {
         <Route path={AppRoutes.Editor} element={<RoutedEditor />} />
       </Routes>
 
-      {state?.backgroundLocation && (
+      {state?.backgroundLocation && state?.shareModalProps && (
         <Routes>
           <Route
             path={AppRoutes.ShareFromList}
-            element={<ShareModal onClose={() => navigate(-1)} />}
+            element={
+              <ShareModal
+                onClose={() => navigate(-1)}
+                {...state.shareModalProps}
+              />
+            }
           />
         </Routes>
       )}
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -1,12 +1,12 @@
 /* global cozy */
 import React, { useState, useEffect, useMemo } from 'react'
 import {
-  useNavigate,
+  HashRouter,
   Route,
   Routes,
   useLocation,
-  useParams,
-  BrowserRouter
+  useNavigate,
+  useParams
 } from 'react-router-dom'
 
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
@@ -127,7 +127,7 @@ const App = ({ isPublic }) => {
 
   return (
     <>
-      <BrowserRouter>
+      <HashRouter>
         <Layout monoColumn={true}>
           {!isPublic && isMobile && (
             <BarCenter>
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -3153,6 +3153,13 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
+"@babel/runtime@^7.7.6":
+  version "7.18.6"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.18.6.tgz#6a1ef59f838debd670421f8c7f2cbb8da9751580"
+  integrity sha512-t9wi7/AW6XtKahAe20Yw0/mMljKq0B1r2fPdvaAdV/KPDZewFXdaaa6K7lxmZBZ8FBNpCiAT6iHPmd6QO9bKfQ==
+  dependencies:
+    regenerator-runtime "^0.13.4"
```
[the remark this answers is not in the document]
acezard: not sure about it, didn't mean to at least. I will review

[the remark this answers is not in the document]
acezard: it must be from some of their linked babel dependencies https://github.com/remix-run/react-router/blob/main/package.json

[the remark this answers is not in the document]
acezard: Must be this one:
```
=> Found "react-virtualized#@babel/runtime@7.14.8"
info This module exists because "cozy-editor-core#react-virtualized" depends on it.
info Disk size without dependencies: "788KB"
info Disk size with unique dependencies: "832KB"
info Disk size with transitive dependencies: "832KB"
info Number of shared dependencies: 1
```

src/components/notes/List/NoteRow.jsx
```diff
@@ -137,16 +136,13 @@ const NoteRow = ({ note, f, t, client }) => {
...
-                modalProps: {
-                  document: { ...note, name: note.attributes.name },
-                  documentType: DocumentTypes.Files,
-                  onClose: history.goBack,
-                  sharingDesc: note.attributes.name
-                }
+            to={AppRoutes.ShareFromList}
+            state={{
+              backgroundLocation: location,
+              modalProps: {
```
acezard, line 142: maybe rename to `shareModal`

src/components/notes/List/index.jsx
```diff
@@ -55,6 +56,8 @@ const ListView = () => {
               </BarRight>
             </WithBreakpoints>
           )}
+
+          <Outlet />
```
acezard, line 60: Maybe clarify why this is used because it's not obvious at all

src/constants/strings.ts
```diff
@@ -41,5 +41,6 @@ export enum Slugs {
 export const SHARING_LOCATION = '/preview/'
 
 export enum DocumentTypes {
```
acezard, line 43: I guess this should be exported from cozy-client instead, what do you think @Crash-- ?

src/components/app.jsx
```diff
@@ -1,11 +1,12 @@
 /* global cozy */
```
acezard, line 1: This is not great should be elsewhere

src/components/app.jsx
```diff
@@ -23,38 +24,40 @@ import { useClient } from 'cozy-client'
...
+        <Route path={AppRoutes.Root} element={<List />} />
+
+        <Route path={AppRoutes.Editor} element={<RoutedEditor />} />
+      </Routes>
+
+      {state?.backgroundLocation && (
+        <Routes>
+          <Route
+            path={AppRoutes.ShareFromList}
+            element={<ShareModal onClose={() => navigate(-1)} />}
```
acezard, line 58: maybe it should go back to root with a replace so you can't go back to modal with the back button. Any thought @Crash-- ?

src/components/notes/List/NoteRow.jsx
acezard: Also maybe it would be better if the share modal could fetch itself all these params from the location params (noteId), that would remove the need to juggle stateful data, we shouldn't be doing that

src/components/notes/List/index.jsx
```diff
@@ -1,21 +1,22 @@
...
 import styles from 'components/notes/List/list.styl'
-import { WithBreakpoints } from './WithBreakpoints'
 import { Breakpoints } from 'types/enums'
+import { WithBreakpoints } from './WithBreakpoints'
 
 const shouldDisplayAddButton = (fetchStatus, notes) =>
   fetchStatus === 'loaded' && notes.length > 0
 
 const ListView = () => {
+  // eslint-disable-next-line no-undef
```
acezard, line 19: Should be handled elsewhere

src/constants/routes.ts
```diff
@@ -1,3 +1,5 @@
-export enum Routes {
+export enum AppRoutes {
+  Editor = 'n/:id',
+  Root = '/',
   ShareFromList = 'share'
```
acezard, line 4: shareFromList should be `/share?n=:id`, and the route would get the data by itself, as it should in idiomatic routing

src/components/notes/List/NoteRow.jsx
```diff
@@ -137,16 +136,13 @@ const NoteRow = ({ note, f, t, client }) => {
...
-                  onClose: history.goBack,
-                  sharingDesc: note.attributes.name
-                }
+            to={AppRoutes.ShareFromList}
+            state={{
+              backgroundLocation: location,
+              shareModalProps: {
+                document: { ...note, name: note.attributes.name },
+                documentType: DocumentTypes.Notes,
+                sharingDesc: note.attributes.name
```
trollepierre, line 145: what is happening now, onClose?

acezard:
[Upgrade React-Router (après React 18)](https://trello.com/c/5Nct1ART/4-upgrade-react-router-apr%C3%A8s-react-18)

PR #318, opened 2022-08-05
fix: Do not force `drive` slug when clicking on `BackFromEditing` button

Ldoppea:
`BackFromEditing` button is responsible to redirect the app to the
`returnUrl` given in the URL's searchParams

In previous behaviour, in a Flagship mobile app, the resulting scenario
would be to call `openApp` with the given URL and the `drive` slug

However, the `drive` slug should not be forced. When the note is opened
from `cozy-notes` then the  `returnUrl` is the `cozy-notes` base url.
So instead we should guess the slug from the `returnUrl`

___

Related PR:

- https://github.com/cozy/cozy-client/pull/1217
- https://github.com/cozy/cozy-ui/pull/2206

___

TODO: 

- [x] Upgrade cozy-client after merging https://github.com/cozy/cozy-client/pull/1217
- [x] Upgrade cozy-ui after merging https://github.com/cozy/cozy-ui/pull/2206

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "cozy-realtime": "4.0.5",
     "cozy-scripts": "6.3.8",
     "cozy-sharing": "^4.5.1",
-    "cozy-ui": "^68.2.0",
+    "cozy-ui": "^70.6.0",
     "lodash": "4.17.21",
     "prop-types": "15.8.1",
     "react": "18.2.0",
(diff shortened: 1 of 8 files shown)


trollepierre (approved):
Ok after update of cozy-client and cozy-ui

acezard (approved):

PR #324, opened 2022-08-16
chore(tools): Add Automerge and Dependabot

trollepierre:


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+  - package-ecosystem: "npm"
+    directory: "/"
+    commit-message:
+      # Prefix all commit messages with "fix"
+      prefix: "fix"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+    labels:
+      - "dependencies"
```
Crash--, line 12: I'm not very convinced by dependabot, in general. But it's even worst on this project since we rely on a lot of third party packages because of atlaskit. And since we need specific versions for those packages, I don't think it's a good idea to upgrade the lockfile.

PR #327, opened 2022-08-25
[WIP] feat: Add persisted store feature

Crash--:
This is a test of the new cozy-client api
in order to have a background fetching.

It seems to work nicely! I need to enable
this feature for the flagship app.

I'll see also, what we can do for the UX


src/targets/browser/configureStore.js
```diff
@@ -0,0 +1,21 @@
+/** global __DEVELOPMENT__ **/
```
Crash--, line 1: can be removed now

rezk2ll:
@Crash-- is this still needed?

PR #328, opened 2022-08-30
fix: Rollback react to v16

acezard:
editorcore is not compatible with React18.
As a result we are downgrading react, react-dom and test libs.
We also update runtime code.

This is not a revert, we keep the changes where we remove enzyme


trollepierre (approved):

Crash--:
Can you add a more detailed commit? Like we need to revert because editor & plugins use `unsafe` without the unsafe prefix

PR #332, opened 2022-09-07
fix: Force sync when clicking on back button, in the note editor

JF-Cozy:
to sync the note with Cozy Drive


Ldoppea (approved):
Code and logic look good. But I can't review the "side effects" part.

Crash--:
Commits should be reverted. Editor should handle that on its own. We already have the call to forceSync() within the useEffect. This is what we should fix.

PR #339, opened 2022-11-10
fix: Open note path on the current tab

zatteo:
Note path was opened on another tab, while the note was opened on the current tab. The new behavior is to open the path on the current tab.

### DIFF
--- a/src/components/notes/List/NoteRow.jsx
+++ b/src/components/notes/List/NoteRow.jsx
@@ -90,12 +90,7 @@ const NoteRow = ({ note, f, t, client }) => {
             </WithBreakpoints>
 
             <WithBreakpoints showOn={Breakpoints.Mobile}>
-              <NotePath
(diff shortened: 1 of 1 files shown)


src/components/notes/List/NoteRow.jsx
```diff
@@ -90,12 +90,7 @@ const NoteRow = ({ note, f, t, client }) => {
             </WithBreakpoints>
 
             <WithBreakpoints showOn={Breakpoints.Mobile}>
-              <NotePath
-                drivePath={drivePath}
-                path={note.path}
-                target="_blank"
-                noLink
-              />
+              <NotePath drivePath={drivePath} path={note.path} noLink />
```
zatteo, line 93: Unused `target="_blank"` because we have the `noLink` property.

Crash-- (approved):

PR #344, opened 2022-11-25
fix(HeaderMenu): show back sharing banner

cballevre:
```
### 🐛 Bug Fixes

* Show back the sharing banner

```


Crash--:
Maybe we can do an another PR to add a test? In order to avoid such a regression?

Crash-- (approved):

cballevre:
I wrote this down in my todo list

PR #345, opened 2022-11-28
fix: preview path to match with cozy-sharing

cballevre:
```
### ✨ Features

*

### 🐛 Bug Fixes

* Correct preview path to show correct sharing banner for cozy to cozy sharing

### 🔧 Tech

*

```

My previous PR : https://github.com/cozy/cozy-notes/pull/344 only fix the problem to show the banner. Notes preview wasn't matching with cozy-sharing pathname so we show every time the public banner with this new fix we show also the name of the cozy owner.

### DIFF
--- a/src/targets/browser/index.jsx
+++ b/src/targets/browser/index.jsx
@@ -29,6 +29,7 @@ import {
 } from 'lib/initFromDom'
 
 import flag from 'cozy-flags'
+import { SHARING_LOCATION } from '../../constants/strings'
 
 const manifest = require('../../../manifest.webapp')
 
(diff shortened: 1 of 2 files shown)


src/targets/browser/index.jsx
```diff
@@ -77,7 +77,7 @@ const renderApp = function(appLocale, client, isPublic) {
                     <SharingProvider
                       doctype="io.cozy.files"
                       documentType="Notes"
-                      previewPath="/preview/"
+                      previewPath="/preview"
```
Crash--, line 80: Let's use the `SHARING_LOCATION` var instead?

cballevre: Yes for sure

Crash-- (approved):

PR #346, opened 2022-12-06
feat: add route to create a note

cballevre:
```
### ✨ Features

* add route to create a note

```

This addition is linked to the addition of a FAB button on cozy-home (cc: https://github.com/cozy/cozy-home/pull/1860).

### DIFF
--- a/src/components/notes/new-route.jsx
+++ b/src/components/notes/new-route.jsx
@@ -1,0 +1,47 @@
+import React, { useEffect, useMemo } from 'react'
+import { useClient } from 'cozy-client'
+import { createNoteDocument, generateReturnUrlToNotesIndex } from 'lib/utils'
+import { generateUniversalLink } from 'cozy-ui/transpiled/react/AppLinker/native'
+import Spinner from 'cozy-ui/transpiled/react/Spinner'
+import get from 'lodash/get'
+import { generateWebLink, useCapabilities } from 'cozy-client'
+import { isFlagshipApp } from 'cozy-device-helper'
+import { useWebviewIntent } from 'cozy-intent'
+
+const NewRoute = () => {
+  const client = useClient()
+  const { capabilities } = useCapabilities(client)
+  const isFlatDomain = useMemo(() => {
+    return get(capabilities, 'flat_subdomains')
+  }, [capabilities])
+  const isFlagship = isFlagshipApp()
+  const webviewIntent = useWebviewIntent()
+
+  useEffect(() => {
+    const createNote = async () => {
+      const { data: doc } = await createNoteDocument(client)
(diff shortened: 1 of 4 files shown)


zatteo (approved):

src/components/notes/new-route.jsx
```diff
@@ -0,0 +1,34 @@
...
+        pathname: `/`
+      })
+      window.location.href = await generateReturnUrlToNotesIndex(
+        client,
+        doc,
+        returnUrl
+      )
+    }
+    createNote()
+  })
```
Crash--, line 29: useEffect without an dep array? 🤔

src/components/notes/new-route.jsx
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    const createNote = async () => {
+      const { data: doc } = await createNoteDocument(client)
+      const returnUrl = generateWebLink({
+        slug: 'notes',
+        cozyUrl: client.getStackClient().uri,
+        subDomainType: isFlatDomain ? 'flat' : 'nested',
+        pathname: `/`
+      })
+      window.location.href = await generateReturnUrlToNotesIndex(
```
Crash--, line 22: Hmm, I think we should not directly change the href. 

Let's do the same we it's already done in Drive: https://github.com/cozy/cozy-drive/blob/953e2c4e25e0c8461b9813f18599d9aa39d38bd2/src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx#L75-L85

src/components/notes/new-route.jsx
```diff
@@ -0,0 +1,34 @@
...
+  const { capabilities } = useCapabilities(client)
+  const isFlatDomain = get(capabilities, 'flat_subdomains')
+
+  useEffect(() => {
+    const createNote = async () => {
+      const { data: doc } = await createNoteDocument(client)
+      const returnUrl = generateWebLink({
+        slug: 'notes',
+        cozyUrl: client.getStackClient().uri,
+        subDomainType: isFlatDomain ? 'flat' : 'nested',
```
Crash--, line 19: `useCapabilities` is async. So you need to wait the value before calling the createNote() method. If not, isFlatDomain will always be undefined

src/lib/utils.js
```diff
@@ -30,11 +30,11 @@ export function getReturnUrl() {
  * @param {object} doc - couchdb document for the note/file
  * @return {string} URL where one can edit the note
  */
-export async function generateReturnUrlToNotesIndex(client, doc) {
+export async function generateReturnUrlToNotesIndex(client, doc, returnUrl) {
```
Crash--, line 33: you should update the jsdoc 🙏

Crash-- (approved):
Looks better :)

PR #358, opened 2022-12-29
feat: Use token from body instead of sharecode in URL

zatteo:
Shared note preview writable again because token from body had write access.

```
### ✨ Features

* Shared note previews are now writable if write access is granted.
```

### DIFF
--- a/src/targets/browser/index.jsx
+++ b/src/targets/browser/index.jsx
@@ -121,8 +121,8 @@ export const initApp = () => {
   const protocol = window.location ? window.location.protocol : 'https:'
 
   const shareCode = getPublicSharecode()
-  const token = shareCode || data.token
-  const isPublic = Boolean(shareCode || !token || token == '')
+  const token = data.token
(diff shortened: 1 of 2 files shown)


src/targets/browser/index.jsx
```diff
@@ -121,8 +121,8 @@ export const initApp = () => {
   const protocol = window.location ? window.location.protocol : 'https:'
 
   const shareCode = getPublicSharecode()
-  const token = shareCode || data.token
-  const isPublic = Boolean(shareCode || !token || token == '')
+  const token = data.token
```
zatteo, line 124: @nono So we can confirm that a token in the body is present in all cases ? I tested successfuly sharing preview write read and read/write access.

nono: Yes, I confirm

Crash-- (approved):

Ldoppea (approved):

PR #367, opened 2023-01-24
feat: Split Atlaskit into a specific chunk

cballevre:
```
### ✨ Features

* Split Atlaskit into a specific chunk
```


Crash--:
Nice! Let's merge it. 

Can you share an image of a webpack stats now?

PR #371, opened 2023-01-30
chore: Add specific build for public page

cballevre:
```
### 🔧 Tech

* Add specific build for public page
```

**Related PRs :**
- https://github.com/cozy/create-cozy-app/pull/1492


src/targets/public/index.ejs
```diff
@@ -0,0 +1,44 @@
...
+    />
+    <link
+      rel="icon"
+      type="image/png"
+      href="./favicon-16x16.png"
+      sizes="16x16"
+    />
+    <link rel="manifest" href="/manifest.json" crossorigin="use-credentials" />
+    <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#297EF2" />
+    <% if (__TARGET__ !== 'mobile') { %>
```
Crash--, line 21: You can remove it

src/targets/public/index.jsx
```diff
@@ -0,0 +1,130 @@
...
+const init = () => {
+  const { appLocale, client } = memoizedInit()
+  renderApp(appLocale, client)
+}
+document.addEventListener('DOMContentLoaded', init)
+
+if (module.hot) {
+  memoizedInit()
+  module.hot.accept()
+}
```
Crash--, line 130: Why do you have copy / pasted this thing, can't we use the other index.jsx?

cballevre: @Crash-- to get an entrypoint so webpack generate an bundle for public which wasn't the case before. I agree with you that we can remove `module.hot` but I will check if removing hot reload doesn't have any impact in watch mode on public page development.

cballevre:
@Crash-- can I have new review on this PR ? pls

Crash-- (approved):

PR #381, opened 2023-06-15
fix: Escape returnURL parameter

Crash--:
Quick fix 

```
### 🔧 Security

* It was possible to inject style in this parameter. CSP was blocking javascript, so this was not too dangerous. 

```

We should rethink this returnURL parameter to see if it still needed or not and maybe check the domain. This param is used at least on Drive when creating a note, in order to go back to Drive when clicking on "Back" and not returning at the root of cozy-notes.

### DIFF
--- a/src/components/notes/editor-loading-error.jsx
+++ b/src/components/notes/editor-loading-error.jsx
@@ -1,26 +1,28 @@
 import React from 'react'
 import { Empty } from 'cozy-ui/transpiled/react'
-import { translate } from 'cozy-ui/transpiled/react/I18n'
+import Button from 'cozy-ui/transpiled/react/Buttons'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
 function EditorLoadingError(props) {
-  const { t, returnUrl } = props
-
+  const { returnUrl } = props
+  const { t } = useI18n()
   return (
     <Empty
       icon="cross-small"
       title={t(`Error.loading_error_title`)}
       text={
-        <p
-          className="u-mb-half"
-          dangerouslySetInnerHTML={{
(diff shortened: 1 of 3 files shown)


src/components/notes/editor-loading-error.jsx
```diff
@@ -1,26 +1,28 @@
 import React from 'react'
 import { Empty } from 'cozy-ui/transpiled/react'
-import { translate } from 'cozy-ui/transpiled/react/I18n'
+import Button from 'cozy-ui/transpiled/react/Buttons'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
 function EditorLoadingError(props) {
-  const { t, returnUrl } = props
-
+  const { returnUrl } = props
```
cballevre, line 7: nit: destructuring the props directly in the function header instead of props

src/components/notes/editor-loading-error.jsx
```diff
@@ -1,26 +1,28 @@
...
-          className="u-mb-half"
-          dangerouslySetInnerHTML={{
-            __html: returnUrl
-              ? t(`Error.loading_error_text_returnUrl`, { url: returnUrl })
-              : t(`Error.loading_error_text_noReturnUrl`)
-          }}
-        />
+        <>
+          <p className="u-mb-half">
+            {t(`Error.loading_error_text_noReturnUrl`)}
```
Ldoppea, line 16: Nit: Should we rename this as we use it also when we have an url?

acezard (approved):

cballevre (approved):

Ldoppea (approved):
Can you add more context about where this url come from?

PR #403, opened 2023-11-27
feat: Try to update @cozy/editor-core

cballevre:


cballevre:
We haven't managed to get the merge up stream to work. For the moment, the subject is on standby.

PR #413, opened 2023-12-11
fix(TrashedBanner): Improve mobile display by using an Alert

cballevre:
Banner component from cozy-ui had wrong margin in mobile. As it will be depreceated soon, we decided to go for Alert which has a similar design but without margin problem.

```
### 🐛 Bug Fixes

* Improve mobile display by using an Alert
```


cballevre:
This is comment test for Linear workflow

Crash-- (approved):

PR #415, opened 2023-12-14
fix(TrashedBanner): Improve feedback to the user

cballevre:
This PR brings real time to the header part with `io.cozy.files`. It also use hook from `cozy-client` to use cache. 

```
### 🐛 Bug Fixes

* Improve feedback to the user on TrashedBanner
```

**Remaining work**
- ~~Check DestroyConfirm design with @joel-costa~~
- Problem the bar dispearing after permanently deletion before the redirection


src/components/notes/editor.jsx
```diff
@@ -92,6 +98,16 @@ export default function Editor(props) {
...
 
+  const [isTrashed, setTrashed] = useState(false)
+
+  useEffect(() => {
+    setTrashed(!!doc?.file?.attributes?.trashed)
+  }, [doc])
+
+  const handleRestore = () => {
+    setTrashed(false)
+  }
```
Crash--, line 109: I don't like the idea of having to handle manually the trashed attribute. Specially since `useNote` is handling it. 

I think you've done that because the `useNote` do not update by itself after the call on `await client.collection('io.cozy.files').restore(noteId)`, right? 

It is not updated because calling the collection directly bypass the cozy-client redux store. 

So we have 3 choices: 
- Call a callback => The choice you have done, but now we have several source for the data and one can be outdated. => KO.
- Expose a client.XXX in order to not bypass the redux store of cozy-client. => Let's see if we can do something about it
- Call a dispatch manually after the call to `restore()`. Like what we do in the `RealTimeQuery` component. 
```
const updatedNote = await client.collection('io.cozy.files').restore(id)
client.dispatch(
    receiveMutationResult(
      client.generateRandomId(),
      updatedNote,
      {},
     Mutations.updateDocument(data)
    )
  )
  ```
  
By doing something like that, the `useNote` should be update to date

Comment (cballevre) [reply src/components/notes/editor.jsx:109]: Me too, but as you guesse, the document doesn't update automatically. I like your proposal, but useNote doesn't use the cozy-client redux store to retrieve information, it retrieves it directly from the stack client ([code](https://github.com/cozy/cozy-notes/blob/3b2fd1e6f109a3e262e6d6de179c039d546f0887/src/lib/collab/stack-client.js#L276)). I'll see if I can't do it some other way to get the updated information.

File: src/components/notes/TrashedBanner/TrashedBanner.jsx [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+  const restore = async () => {
+    try {
+      await client.collection('io.cozy.files').restore(noteId)
+      showAlert(t('TrashedBanner.restoreSuccess'), 'secondary')
+    } catch {
+      showAlert(t('TrashedBanner.restoreError'), 'error')
+    } finally {
+      setBusy(false)
+    }
+  }
```
Crash--, line 35: We can't use `useMutation` because of the need to call a specific method of the collection (restore)?

test/AppLike.jsx
```diff
@@ -1,12 +1,23 @@
...
+
+const AppLike = ({ children, client }) => {
+  return (
+    <CozyProvider client={client}>
+      <TestI18n>{children}</TestI18n>
+    </CozyProvider>
+  )
+}
+
+export { TestI18n, AppLike }
```
Crash--, line 23: What is the goal and benefit? What is the cost to use AppLike everywhere?

src/components/notes/editor.jsx
```diff
@@ -168,7 +157,7 @@ export default function Editor(props) {
                 collabProvider.serviceClient.cozyClient.getStackClient().uri
               }
               isPublic={isPublic}
-              file={doc.file}
+              file={file}
```
Crash--, line 150: Should we render all these components only if `file` is defined? Having to check `if(file)` everywhere (back_from_editing, header_menu, headernote...) doesn't seem right.

src/hooks/useUser.js
```diff
@@ -1,15 +1,15 @@
...
     () =>
-      providedUserName ||
-      getUserNameFromUrl() ||
-      (isPublic ? '?' : getShortNameFromClient(cozyClient)),
-    [cozyClient, providedUserName, isPublic]
+      getUserNameFromUrl() || (isPublic ? '?' : getShortNameFromClient(client)),
+    [client, isPublic]
   )
   const userId = userName
   return { userId, userName }
```
Crash--, line 15: (nit: since this code rely on hook & context so React, should be `jsx`. To me `.js` should be framework agnostic)

src/components/notes/editor.jsx
```diff
@@ -42,10 +42,7 @@ export default function Editor(props) {
 
   // plugins and config
   const isPublic = useContext(IsPublicContext)
-  const { userName, userId } = useUser({
-    userName: props.userName,
-    cozyClient
-  })
+  const { userName, userId } = useUser()
```
Crash--: Hmm, `props.userName` was never set?

src/lib/collab/stack-client.js
```diff
@@ -275,25 +275,35 @@ export class ServiceClient {
...
+    return formatedDoc
+  }
+
+  /**
+   * Get the full document for a note
+   * @param {uuid} noteId
+   */
+  getFormatedDoc(note) {
+    const formatedDoc = this.formatDoc(note)
     this.schemaVersion = formatedDoc.schemaVersion
```
Crash--, line 289: Are we sure about this sideEffect (setting `this.schemaVersion`)?

src/components/notes/TrashedBanner/TrashedBanner.jsx
cballevre: Unfortunately, the useMutation API only allows the use of a save and not to pass a function as in its react-query equivalent. Perhaps this is something that needs to be rethink. We could transform the current one into `useSave` and keep `useMutation` for more generic cases like this one

test/AppLike.jsx
cballevre: The aim was to have a lighter wrapper to match with other apps, but I've never done a comparison to find out if there was better performance

src/components/notes/editor.jsx
cballevre: For me Editor is only used into 2 places : 
- https://github.com/cozy/cozy-notes/blob/3b2fd1e6f109a3e262e6d6de179c039d546f0887/src/components/app.jsx#L39
- https://github.com/cozy/cozy-notes/blob/3b2fd1e6f109a3e262e6d6de179c039d546f0887/src/components/app.jsx#L105
No of them shared an userName property but I can be wrong

src/lib/collab/stack-client.js
cballevre: I only follow the [getDoc](https://github.com/cozy/cozy-notes/blob/3b2fd1e6f109a3e262e6d6de179c039d546f0887/src/lib/collab/stack-client.js#L276) function but without the stackClient request

src/components/notes/editor.jsx
cballevre: You are right its remainder from a previous development iteration

src/hooks/useUser.js
cballevre: I renamed all the hooks that had React to jsx

src/components/notes/editor.jsx
cballevre: I refactored this part to remove the conditions

Crash-- (approved):

PR #421, opened 2024-01-24
fix: Allow copy pasting images from note to note (VO-125)

acezard:
### ✨ Features

* It will now be possible to copy an image from an note to another one provided it is the same cozy instance owning both notes


cballevre (approved):
The code LGTM 👍

PR #429, opened 2024-04-29
fix: make toolbar stuck to bottom with ios (VO-704)

acezard:
### 🐛 Bug Fixes

* Fixed toolbar position on iOS devices

https://github.com/cozy/cozy-notes/assets/12577784/17af8cf2-a9d2-4a28-b98f-c26d426cdc33

### DIFF
--- a/src/hooks/useFixedToBottomOnIos.ts
+++ b/src/hooks/useFixedToBottomOnIos.ts
@@ -1,0 +1,44 @@
+import { useEffect } from 'react'
+
+import {
+  applyFixToElement,
+  handleMutation,
+  setupEventListeners,
+  tearDownEventListeners,
+  TOOLBAR_SELECTOR
+} from 'lib/patches/fixedToBottomIos'
+
+/**
+ * Custom hook that fixes an element to the bottom of the screen on iOS devices.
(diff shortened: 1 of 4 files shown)


src/hooks/useFixedToBottomOnIos.ts
```diff
@@ -0,0 +1,44 @@
...
+    const existingElement = document.querySelector<HTMLElement>(
+      TOOLBAR_SELECTOR
+    )
+
+    if (existingElement) {
+      applyFixToElement(existingElement)
+      setupEventListeners(existingElement)
+    } else {
+      observer = new MutationObserver(handleMutation)
+      observer.observe(document.body, { childList: true, subtree: true })
```
acezard, line 34: we could use the element react ref, but that would mean digging into atlaskit. Here I chose to not do that. But that would be possible. It was mostly an arbitrary decision

cballevre: I find that it makes it more complicated, given that we're moving away from React, but I understand that you want to get as far away from Atlaskit as possible

cballevre (approved):
I don't see anything that would prevent to merge as the fix has restricted scope

PR #468, opened 2025-04-15
feat: Update app icons

JF-Cozy:


src/targets/vendor/assets/safari-pinned-tab.svg
zatteo, line 1: Strange no ?

zatteo (approved):

src/targets/vendor/assets/safari-pinned-tab.svg
JF-Cozy: changed

